PR #29, opened 2022-09-12
Fix formating in Backup Rancher

MaximilianKindshofer:
Delete a space so the warning renders correctly - tiny change... keep up the great work!


btat (approved):
Thanks for catching this.

PR #40, opened 2022-09-13
Add i18n and Chinese translation

vickyhella:


btat:
There are no errors when I build this, but whenever I try to switch to the Chinese docs I get the 'Page Not Found' page.

btat (changes requested):
There are no errors when I build this, but whenever I try to switch to the Chinese docs I get the 'Page Not Found' page.

vickyhella:
May I ask did you run `yarn start`? If so, `yarn start` does not work for the non-default locales. As per Docusaurus's Documentation, you can use `yarn run start -- --locale zh` to run the `zh` locale specifically.

I have also temporarily deployed the vk-bilingual branch to Netlify at https://vk-rancher.netlify.app/. You can visit it to see how it looks in production if needed.

image

btat (approved):

PR #219, opened 2022-09-20
Add workflow for stale issues/PRs

btat:
Uses the config from [rancher/rancher](https://github.com/rancher/rancher/blob/release/v2.7/.github/workflows/stale.yml) as a reference.

Slight modifications include:
- 90 days before marking an item as stale
- 30 days before items marked stale are closed
- Only items labeled `internal` are exempt. Most of the other labels do not exist in our docs repo. We can revisit this later


gunamata (approved):
LGTM

PR #248, opened 2022-10-10
Add note about resource group and vnet

dkeightley:
Adding note about prefix required for the VNet when in a different RG from the VMs (node template).

Reference https://github.com/rancher/rancher/issues/12582#issuecomment-424989612

### DIFF
--- a/docs/reference-guides/cluster-configuration/downstream-cluster-configuration/node-template-configuration/azure.md
+++ b/docs/reference-guides/cluster-configuration/downstream-cluster-configuration/node-template-configuration/azure.md
@@ -10,6 +10,12 @@ Account access information is stored as a cloud credential. Cloud credentials ar
 - **Network** configures the networking used in your cluster.
 - **Instance** customizes your VM configuration.
 
+:::note
+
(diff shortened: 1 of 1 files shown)


docs/reference-guides/cluster-configuration/downstream-cluster-configuration/node-template-configuration/azure.md
```diff
@@ -10,6 +10,12 @@ Account access information is stored as a cloud credential. Cloud credentials ar
 - **Network** configures the networking used in your cluster.
 - **Instance** customizes your VM configuration.
 
+:::note
+
+If using a VNet in a different Resource Group than the VMs, the VNet name should be prefixed with with the Resource Group name, example: `<resource group>:<vnet>`
```
btat, line 15:
```suggestion
If using a VNet in a different Resource Group than the VMs, the VNet name should be prefixed with the Resource Group name. For example, `<resource group>:<vnet>`.
```

btat (changes requested):

PR #282, opened 2022-11-02
Update Azure AD documentation, change permission requirements

maxsokolovsky:
The primary change is the new suggested permissions for the Azure app (only `Directory.Read.All` is suggested).

Other changes aim to make the instructions a little clearer in a few points. The screenshots are also updated to reflect the UI changes.


maxsokolovsky:
The global version selector is still v2.6, we would need to update it for v2.7.

Also, is there a way to make the navbar on the right work with tabs?

btat:
> The global version selector is still v2.6, we would need to update it for v2.7.

Thanks for the catch, we'll get that updated.

> Also, is there a way to make the navbar on the right work with tabs?

I don't believe there's a solution for the navbar+tabs combo quite yet as we haven't been able to find anything in the Docusaurus docs. But we'll it's something we're aware of.

Also, are there any reviewers you'd like me to add? I don't think EIO has gotten to our request to allow engineers to add reviewers.

maxsokolovsky:
We'd need to also update this v2.6 badge above the header on each page.

Sure, could you add @jtravee and @MbolotSuse? Thanks!

jtravee (approved):

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -104,65 +103,60 @@ Next, set API permissions for Rancher within Azure.
...
 
-1. Your endpoint options will typically be [Standard](#global) and [China](#china). With these options, you need only enter the **Tenant ID**, **Application ID**, and **Application Secret** (Rancher will take care of the rest).
+1. Your endpoint options will typically be either [Standard](#global) or [China](#china). With these options, you only enter the **Tenant ID**, **Application ID**, and **Application Secret**.
 
 ![Standard Endpoint Options](/img/tenant-application-id-secret.png)
 
 **For Custom Endpoints:**
 
-**Warning:** Custom Endpoints are not supported nor fully tested by Rancher.
+**Warning:** Custom Endpoints are not tested and thus not fully supported by Rancher.
```
MbolotSuse, line 159: Nit: I would rephrase this  as `Custom Endpoints are not tested or fully supported by Rancher`. The way it is currently written, it implies that if they were tested then they would be supported, which I don't think we want to imply.

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -218,14 +212,17 @@ The following table maps the custom config values you copied in the Azure portal
 
 ### Migrating from Azure AD Graph API to Microsoft Graph API
 
-Since [Azure AD Graph API](https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-overview) was deprecated in June 2022 and will be retired at the end of 2022, users should update their Azure AD App to use the new [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/use-the-api) in Rancher.
+Since [Azure AD Graph API](https://docs.microsoft.com/en-us/graph/migrate-azure-ad-graph-overview) was deprecated in June 2022 and will be retired at the end of 2022, admins should update their Azure AD App to use the new [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/use-the-api) in Rancher. They should do so as soon as possible. Otherwise, users may not be able to log in to Rancher via Azure AD.
```
MbolotSuse, line 215: I would change:

> They should do so as soon as possible. Otherwise, users may not be able to log in to Rancher via Azure AD.

to something like:

> This should be done well in advance of the endpoint being retired. If Rancher is still configured to use the old Azure AD Graph API when it is retired, users may not be able to log into Rancher using Azure AD.

MbolotSuse (changes requested):
Got a couple of changes on the non-images file (I still need to run this locally so I can see what the whole thing looks like in context)

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -6,7 +6,7 @@ import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
 <Tabs>
-  <TabItem value="Rancher v2.6.7">
+  <TabItem value="Rancher v2.7.0+">
```
MbolotSuse, line 9: The flow here with this change is kind of odd. There's now two tabs - one for v2.6.0-v2.6.6 and one for v2.7.0+. It kind of makes it seem like azureAD is not supported on 2.6 versions after 2.6.6. Can you elaborate on why this change was made?

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -83,16 +81,17 @@ From the Azure portal, create a client secret. Rancher will use this key to auth
 
      ![Open Rancher Registration](/img/open-rancher-app-reg.png)
 
-1. From the navigation pane on left, click **Certificates and Secrets**.
+1. From the navigation pane on left, click **Certificates & secrets**.
```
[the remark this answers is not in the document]
maxsokolovsky: I was just following the actual text in the Azure portal as it appears today.

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -83,16 +81,17 @@ From the Azure portal, create a client secret. Rancher will use this key to auth
...
 
 1. Click **New client secret**.
     ![Create new client secret](/img/new-client-secret.png)
 1. Enter a **Description** (something like `Rancher`).
 1. Select duration for the key from the options under **Expires**. This drop-down sets the expiration date for the key. Shorter durations are more secure, but require you to create a new key after expiration.
+Note that Rancher will disable Azure AD when it detects that the secret has expired, so rotate and update it in Rancher before it expires.
```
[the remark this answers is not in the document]
maxsokolovsky: One customer mentioned that Rancher automatically disabled their auth config when the application secret expired. I haven't seen this myself, will try reproducing this now to know for sure.

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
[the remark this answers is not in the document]
MbolotSuse: Ah ok - if this is character for character identical to the azure portal feel free to leave as is.

MbolotSuse:
Some comments on the images (since I can't comment on them directly):

- `static/img/azure-update-popup.png`: I don't think that the `Directory.Read.All Application` is the right thing to highlight. By this point, the users should have already enabled these permissions. Can you elaborate on why you thought this line would be better to highlight?
- `static/img/rancher-button-to-update.png`: I think that having something to point out where the "Update Endpoint" button was a good idea - can you elaborate more on why you removed the box (I do think an arrow would work better in this scenario)?
- `static/img/application-client-id.png` - this appears to be the same as `static/img/app-configuration.png` - do we need both of these images?

MbolotSuse (changes requested):
Second pass on images/flow

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
maxsokolovsky: Mostly because these docs are v2.7 docs, so any v2.6 references are not relevant for it.

MbolotSuse: From my side, I see these docs with a drop-down of 2.6 at the top and there's another tab for `Rancher v2.6.0 - v2.6.6`. Given what you have said here, I would expect to only see stuff for 2.7 - is this intentional?

maxsokolovsky: This is known - either the docs team have not yet updated it, or they have, and I just haven't rebased against a more recent version of the branch.

maxsokolovsky:
> Some comments on the images (since I can't comment on them directly):
> 
> * `static/img/azure-update-popup.png`: I don't think that the `Directory.Read.All Application` is the right thing to highlight. By this point, the users should have already enabled these permissions. Can you elaborate on why you thought this line would be better to highlight?
> * `static/img/rancher-button-to-update.png`: I think that having something to point out where the "Update Endpoint" button was a good idea - can you elaborate more on why you removed the box (I do think an arrow would work better in this scenario)?
> * `static/img/application-client-id.png` - this appears to be the same as `static/img/app-configuration.png` - do we need both of these images?

1. In case admins click the button first and don't yet update the permissions.
2. We do specify the button to be clicked above. I also don't know how to make that arrow. :)
3. I removed the reference to the older image, can't delete it fully, since it's used by the docs on the v2.6 branch.

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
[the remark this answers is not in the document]
maxsokolovsky: You're right - I checked and saw that on secret expiration there is just an error on login and in the logs. It says the secret has expired. The provider is not disabled. Will adjust the text.

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
MbolotSuse: I think that this makes for a confusing user experience. If we can fix this, I would recommend doing so. If these docs are also intended for 2.7.0+, I think that it makes sense to not have a tab for 2.7.0+ - it's already on the 2.7 page, so users know that it is for 2.7.

maxsokolovsky: @btat, what do you think? I am also not sure how to remove a tab item without breaking the page...

MbolotSuse (changes requested):
Some minor suggestions

docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-azure-ad.md
```diff
@@ -83,94 +82,83 @@ From the Azure portal, create a client secret. Rancher will use this key to auth
...
      ![Open Rancher Registration](/img/open-rancher-app-reg.png)
 
-1. From the navigation pane on left, click **Certificates and Secrets**.
+1. From the navigation pane on left, click **Certificates & secrets**.
 
 1. Click **New client secret**.
     ![Create new client secret](/img/new-client-secret.png)
 1. Enter a **Description** (something like `Rancher`).
 1. Select duration for the key from the options under **Expires**. This drop-down sets the expiration date for the key. Shorter durations are more secure, but require you to create a new key after expiration.
+Note that users won't be able to log in to Rancher when it detects that the application secret has expired, so rotate the secret in Azure and update in Rancher before it expires.
```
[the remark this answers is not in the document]
maxsokolovsky: I like these special notes and warnings, but I am trying to avoid excessive formatting, which can make the docs too dense.
There is so much going on already with bold, italic, underscores, links, etc.

MbolotSuse (approved):

MbolotSuse:
I think there's still improvements that we could make to these docs, but I'll approve in the interest of getting something better in that describes the new permissions.

btat (approved):

PR #284, opened 2022-11-02
Add docs for scaling tips

rmweir:
https://github.com/rancher/rancher/issues/37983

### DIFF
--- a/docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
+++ b/docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
@@ -1,0 +1,61 @@
+---
+title: Tips for Scaling Rancher
+---
+
+This guide aims to introduce the approaches that should be considered to scale Rancher setups, and associated challenges with doing so. As systems grow performance will naturally reduce, but there are steps we can take to minimize the load put on Rancher, as well as optimize Rancher's ability to handle these larger setups.
+
+## General Tips on Optimizing Rancher's Performance
+* It is advisable to keep Rancher up to date with patch releases. Performance improvements and bug fixes are made throughout the life of a minor release. You can review the release notes to help inform your own decisions on whether an upgrade is necessary but we recommend keeping yourself up to date in most cases.
+
+* Performance will be negatively impacted by increased latency between Rancher's infrastructure and a downstream cluster's infrastructure (eg. geographic distance). If a user or organization requires clusters/nodes all over the world or spread across many regions, it is best to use multiple Rancher installations.
+
+* Please always try to scale up gradually, monitoring and observing any change in behavior while doing do. It is usually easier to resolve performance problems as soon as they surface, and before other problems confuse symptoms. 
+
+## Minimizing Load on the local cluster
+The largest bottleneck when scaling Rancher is resource growth in the local Kubernetes cluster. The local cluster contains information for all downstream clusters. Many operations that apply to downstream clusters will create new objects in the local cluster and require computation from handlers running in the local cluster.
+
+### Managing Your Object Counts
+ETCD eventually encounters limitations to the number of a single Kubernetes resource type it can store. These exact numbers are not well documented. From internal observations we usually see performance issues once a single resource type's object count exceeds 60k, and often that type is Rolebindings.
+
+Rolebindings are created in the local cluster as a side effect of many operations.
+
+Considerations when attempting reduce rolebindings in the local cluster:
+* Only add users to clusters and projects when necessary
+* Remove clusters and projects when they are no longer needed
+* Only use custom roles if necessary
+* Use as few rules as possible in custom roles
+* Consider whether adding a role to a user is redundant
+* Consider that using less, but more powerful, clusters may be more efficient
+* Experiment to see if creating new projects or creating new clusters manifests in fewer rolebindings for your specific use case.
+
+### Using New Apps Over Legacy Apps
+There are two app kubernetes resources that Rancher uses: apps.projects.cattle.io and apps.cattle.cattle.io. The legacy apps, apps.projects.cattle.io, were introduced first in the Cluster Manager and are now outdated. The new apps, apps.catalog.cattle.io, are found in the Cluster Explorer for their respective cluster. The new apps are preferrable because they live in the downstream cluster while the legacy apps live in the local cluster.
+
+We recommend removing apps that appear in the Cluster Manager, replacing them with apps in the Cluster Explorer for their target cluster if necessary and creating any future apps in the cluster's Cluster Explorer only.
+
+### Using the Authorized Cluster Endpoint (ACE)
+There is an _Authorized Cluster Endpoint_ option for Rancher provisioned RKE1, RKE2, and K3s clusters. When enabled this adds a context to kubeconfigs generated for the cluster that uses a direct endpoint to the cluster and bypasses Rancher. However, it is not enough to only enable this option. The user of the Kubeconfig needs to use `kubectl use-context <ace context name>` in order to start using it.
+
+Without using ACE, all kubeconfig requests first route through Rancher.
+
+### Experimental: Option to Reduce Event Handler Executions
+The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, and when Rancher is started. Additionally, they run every 15 hours when caches are synced. In scaled setups these scheduled runs come with huge performance costs because every handler is being run on every applicable object. However, this scheduled execution of handlers can be disabled using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help.
+
+The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers. Adding the controller types to the variable will disable that set of controllers from running their handlers as part of cache resyncing.
+
+* `mgmt` refers to management controllers which only run on one Rancher node.
+* `user` refers to user controllers which run for every cluster. Some of these are ran on the same node as management controllers, while other run in the downstream cluster. This will option targets the former.
+* `scaled` refers to scaled controllers which run on every Rancher node. This is not recommended to be set due to the critical functionality the scaled handlers are responsible for.
+
+In short, if you notice CPU usage peaks every 15 hours, add the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable to your rancher deployment with the value `mgmt,user`.
+
+## Optimizations Outside of Rancher
+A large component of performance is the local cluster and how it was configured. This cluster can introduce a bottleneck before Rancher software ever runs. When Rancher nodes experience high resource usage, you can use the command "top" to identify whether it is Rancher or a Kubernetes component that is consuming the resource in excess.
+
+### Keeping Kubernetes Versions Up to Date
+Similar to Rancher versions, it is advisable to keep your kubernetes cluster up to date. This will ensure that your cluster contains any available performance enhancements or bug fixes.
+
+### Optimizing ETCD
+The two main bottlenecks to [ETCD performance](https://etcd.io/docs/v3.4/op-guide/performance/) are disk speed and network speed. Optimization to either should improve performance. For information regarding ETCD performance see [Slow etcd performance (performance testing and optimization)](https://www.suse.com/support/kb/doc/?id=000020100) and [Tuning etcd for Large Installations](https://docs.ranchermanager.rancher.io/how-to-guides/advanced-user-guides/tune-etcd-for-large-installs). Information on disks can also be found [in our docs](https://docs.Ranchermanager.Rancher.io/v2.5/pages-for-subheaders/installation-requirements#disks).
+
+Theoretically, the more nodes in an ETCD cluster the slower it will be due to replication requirements [source](https://etcd.io/docs/v3.3/faq). This may be counter-intuitive to common scaling approaches. It can also be inferred that ETCD performance will be inversely affected by distance between nodes as that will slow down network communication.


docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Tips for Scaling Rancher
+---
+
+This guide aims at introducing challenges and approaches that should be considered to scale Rancher setups. As setups grow performance will reduce. There are steps we can take to minimize the load that is put on Rancher and optimize Rancher's ability to handle increased loads.
```
cbron, line 5:
```suggestion
This guide aims to introduce the approaches that should be considered to scale Rancher setups, and associated challenges with doing so. As systems grow performance will naturally reduce, but there are steps we can take to minimize the load put on Rancher, as well as optimize Rancher's ability to handle these larger setups.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+## General Tips on Optimizing Rancher's Performance
+* It is advisable to keep Rancher up to date with patch releases. Performance improvements and bug fixes are made throughout the life of a minor release. You can review the release notes to help inform your own decisions on whether an upgrade is necessary but we recommend keeping yourself up to date in most cases.
+
+* The geographically closer Rancher's infrastructure and downstream cluster's infrastructure are the better. If a user or organization requires clusters/nodes all over the world or spread across many regions, it is best to use multiple Rancher installations.
```
cbron, line 10:
```suggestion
* Performance will be negatively impacted by increased geographic distance between Rancher's infrastructure and a downstream cluster's infrastructure. If a user or organization requires clusters/nodes all over the world or spread across many regions, it is best to use multiple Rancher installations.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Minimizing Load on the local cluster
+The largest bottleneck when scaling Rancher is kubernetes resource growth in the local cluster. The local cluster contains information for all downstream clusters. Many operations that apply to downstream clusters will create new objects in the local cluster and require computation from handlers running in the local cluster.
```
cbron, line 13:
```suggestion
The largest bottleneck when scaling Rancher is resource growth in the local Kubernetes cluster. The local cluster contains information for all downstream clusters. Many operations that apply to downstream clusters will create new objects in the local cluster and require computation from handlers running in the local cluster.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Managing Your Object Counts
+ETCD eventually encounters limitations to the number of a single kubernetes resource type it can store. These exact numbers are not well documented. From internal observations we usually see performance issues once an object count exceeds 60k. Usually the kubernetes resource type that reaches these numbers is Rolebinding.
```
cbron, line 16:
```suggestion
ETCD eventually encounters limitations to the number of a single Kubernetes resource type it can store. These exact numbers are not well documented. From internal observations we usually see performance issues once a single resource type's object count exceeds 60k, and often that type is Rolebindings.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+Rolebindings are created in the local cluster as a side effect of many operations. This includes creating clusters, adding users, adding projects, and adding roles to users. When adding users to cluster or projects, rolebindings will be created.
```
cbron, line 18: I would scratch second half here, you document it well enough in bullets below
```suggestion
Rolebindings are created in the local cluster as a side effect of many operations. 
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+Considerations when attempting reduce rolebindings in the local cluster:
+* Only add users to clusters and projects when necessary
+* Remove clusters and projects when they are no longer needed
+* Only use custom roles if necessary
+* Use as few rules as possible in custom roles
+* Consider whether adding a role to a user is redundant
+* Consider that using less, but more powerful, clusters may be more efficient
```
cbron, line 26:
```suggestion
* Consider that using fewer powerful downstream clusters will be more efficient
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+* Monitoring the rolebindings created from creating new projects for specific purposes versus creating new clusters
```
cbron, line 27: I don't understand this one. Which way are we directing users ?

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Using Apps V2 Over Apps V1
+There are two App kubernetes resources that Rancher uses: apps.projects.cattle.io and correct this apps.cattle.cattle.io. V1 here refers to apps.projects.cattle.io since they were introduced first and are legacy, while V2 refers to apps.catalog.cattle.io. Apps V2 are preferrable because they live in the downstream cluster while Apps V1 lives in the local cluster. The dashboard uses Apps V2 while the old `/g/clusters` UI uses Apps V1.
```
cbron, line 30: You should use `TODO` :) 

```suggestion
There are two Application resources that Rancher uses: apps.projects.cattle.io and apps.cattle.cattle.io. V1 here refers to apps.projects.cattle.io since they were introduced first and are legacy, while V2 refers to apps.catalog.cattle.io. Apps V2 are preferrable because they live in the downstream cluster while Apps V1 lives in the local cluster. The dashboard uses Apps V2 while the old `/g/clusters` UI uses Apps V1.
```

Is this worth including since we are already migrating away from cluster manager ? It would be simpler if we just say "use cluster explorer apps", than listing the CRD types.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Using the Authorized Cluster Endpoint (ACE)
+There is an option for RKE cluster called Authorized Cluster Endpoint. When enabled this adds a context to kubeconfigs generated for the cluster that uses a direct endpoint to the cluster. It is not enough to enable this option. The user of the config needs to use `kubectl use-context <ace context name>` in order to start using it.
```
cbron, line 33:
```suggestion
There is an _Authorized Cluster Endpoint_ option for Rancher provisioned RKE1, RKE2, and K3s clusters. When enabled this adds a context to kubeconfigs generated for the cluster that uses a direct endpoint to the cluster and bypasses Rancher. However, it is not enough to only enable this option. The user of the Kubeconfig needs to use `kubectl use-context <ace context name>` in order to start using it.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+Without using ACE, all kubeconfig requests first route through Rancher. This contributes to load on the Rancher setup, while using ACE completely bypasses the Rancher setup.
```
cbron, line 35:
```suggestion
Without using ACE, all kubeconfig requests first route through Rancher. This contributes to load on the Rancher setup.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Option to Reduce Event Handler Executions
```
cbron, line 37: We should add an experimental italic phrase before the text block, or put it in a section for that

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Optimizations Outside of Rancher
+A large component of performance is the local cluster and how it was configured. Since Rancher is "Bring your own cluster", this impactful component introduces a potential bottleneck before Rancher software ever runs. When Rancher nodes experience high resource usage, you can use the command "top" to identify whether it is Rancher or a kubernetes component that is consuming the resource in excess.
```
cbron, line 41:
```suggestion
A large component of performance is the local cluster and how it was configured. This cluster can introduce a bottleneck before Rancher software ever runs. When Rancher nodes experience high resource usage, you can use the command "top" to identify whether it is Rancher or a Kubernetes component that is consuming the resource in excess.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Using Alternatives to ETCD
+Some kubernetes distributions such as K3S and RKE2 support the use of alternative databases in place of ETCD. Measurements of exact gains are not documented at this time but it is possible that using a data store that is not ETCD would yield performance improvements. Documentation for using alternatives with K3S can be found [here](https://docs.k3s.io/installation/datastore).
```
cbron, line 50: Not supported yet unfortunately, need to remove

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Keeping Kubernetes Versions Up to Date
+Similar to Rancher versions, it is advisable to keep your kubernetes cluster up to date. This will ensure that your cluster contains any available performance enhancements or bug fixes.
+
+### Optimizing ETCD
+As mentioned [here](https://etcd.io/docs/v3.4/op-guide/performance/), the two main bottlenecks to ETCD performance are disk speed and network speed. Optimization to either should improve performance. For information regarding ETCD performance see [Slow etcd performance (performance testing and optimization)](https://www.suse.com/support/kb/doc/?id=000020100). Information on disk can also be found [here](https://docs.Ranchermanager.Rancher.io/v2.5/pages-for-subheaders/installation-requirements#disks). Theoretically, the more nodes in an ETCD cluster the slower it will be due to replication requirements [source](https://etcd.io/docs/v3.3/faq). It can be inferred that ETCD performance will be inversely affected by distance between nodes as that will slow down network communication.
```
cbron, line 47: * Good call including that KB link.
* Break up bottom paragraph
* Using "here" as a link is considered bad UX, "source" is fine though

```suggestion
The two main bottlenecks to [ETCD performance](https://etcd.io/docs/v3.4/op-guide/performance/) are disk speed and network speed. Optimization to either should improve performance. For information regarding ETCD performance see [Slow etcd performance (performance testing and optimization)](https://www.suse.com/support/kb/doc/?id=000020100). Information on disks can also be found [in our docs](https://docs.Ranchermanager.Rancher.io/v2.5/pages-for-subheaders/installation-requirements#disks). 

Theoretically, the more nodes in an ETCD cluster the slower it will be due to replication requirements [source](https://etcd.io/docs/v3.3/faq). This may be counter-intuitive to common scaling approaches. It can also be inferred that ETCD performance will be inversely affected by distance between nodes as that will slow down network communication.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Option to Reduce Event Handler Executions
+The bulk of Rancher's logic are ran by event handlers. These event handlers run on an object whenever the object is updated. They also run when Rancher is started. By default they run every 15 hours when caches are synced but this can be changed using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help. The value for the environment variable can be a comma separated list of the following options: mgmt, user, and scaled. The values refer to types of controllers (the structures that contain and run handlers) and their handlers. Mgmt refers to management which only runs on one Rancher node. User will run for every cluster. Some user handlers run on a Rancher node while others run downstream in their respective cluster. Scaled runs on every Rancher node. The two options that cover most controllers' handlers and are supported are mgmt and user. There are very few "scaled" controllers and their handlers are often critical. It is not recommended that the scaled value is used.
```
cbron, line 38: Re-formatted below, but the first paragraph still doesn't explain explicitly what this setting does. Does it change the 15 hours ? Do they still sync on object update ?  

```suggestion
The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, or when Rancher is started. By default they run every 15 hours when caches are synced, however this can be changed using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help. 

The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers.
* `mgmt` refers to management which only runs on one Rancher node
* `user` will run for every cluster
* `scaled` runs on every Rancher node, this is not recommended to be set due to critical functionality
```

cbron (changes requested):
Lots of editing, but will be good in the end.

cbron:
Once backend is done reviewing we should get a docs reviewer as well

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: I think that the revised version my be interpreted differently then intended. That could mean, use less powerful cluster than you are currently using instead of use less clusters but more powerful ones.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: This is a call for them to track it themselves as I think it depends on the user case. Can remove if you think we should.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: The CRD is literally called App(s). I think Applications would be misleading. I could included that it is short for applications? Also what do you mean by using TODO? Yes I think it's relevant since a lot of scaled users have used rancher for a while and we don't know exactly what they are doing, especially because we have not made a migration path for Apps public.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: I suppose this is fine for now but in the future I would like to dive more into how the user brings the cluster and what that entails if they want a performant system.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir:
```suggestion
The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, and when Rancher is started. Additionally, they run every 15 hours when caches are synced. In scaled setups these scheduled runs come with huge performance costs because every handler is being run on every applicable object. However, this scheduled execution of handlers can be disabled using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help.

The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers.
* `mgmt` refers to management which only runs on one Rancher node
* `user` will run for every cluster
* `scaled` runs on every Rancher node, this is not recommended to be set due to critical functionality the scaled handlers are responsible for
```
These are the changes I've made to hopefully clarify what is happening.

rmweir:
@cbron I have included most of you recommendations and commented on any others. I left a new commit to help with review the new updates. I will squash once you are ready to approve.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
git-ival: Should we make a reference to existing scaling/performance caveats of upstream K8s? You can see this in the files at https://github.com/kubernetes/community/tree/master/sig-scalability/configs-and-limits

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
cbron: Good point

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
cbron: How about: 
```suggestion
* Experiment to see if creating new projects or creating new clusters manifests in fewer rolebindings for your specific use case. 
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
cbron: You had this, which I assumed was a todo: 
>  and correct this apps.cattle.cattle.io

Even if we want to tell people this, do we think giving them the CRD's is the best course of action ? What are you hoping they do with this information, to get a count of both CRD types and see which is larger ? Or just migrate old apps to explorer apps. If we keep this info I would like to rewrite it.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
cbron: I think that'd be good. We often defer to them. Though I would put it up above in the general section.

git-ival (approved):
LGTM, might also be good to reference the scalability docs from upstream when referencing limitations but its fine as-is

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Using Apps V2 Over Apps V1
+There are two App kubernetes resources that Rancher uses: apps.projects.cattle.io and correct this apps.cattle.cattle.io. V1 here refers to apps.projects.cattle.io since they were introduced first and are legacy, while V2 refers to apps.catalog.cattle.io. Apps V2 are preferrable because they live in the downstream cluster while Apps V1 lives in the local cluster. The dashboard uses Apps V2 while the old `/g/clusters` UI uses Apps V1.
+
+### Using the Authorized Cluster Endpoint (ACE)
+There is an _Authorized Cluster Endpoint_ option for Rancher provisioned RKE1, RKE2, and K3s clusters. When enabled this adds a context to kubeconfigs generated for the cluster that uses a direct endpoint to the cluster and bypasses Rancher. However, it is not enough to only enable this option. The user of the Kubeconfig needs to use `kubectl use-context <ace context name>` in order to start using it.
+
+Without using ACE, all kubeconfig requests first route through Rancher.
+
+### Experiemntal: Option to Reduce Event Handler Executions
```
cbron, line 37: typo
```suggestion
### Experimental: Option to Reduce Event Handler Executions
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,54 @@
...
+The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, and when Rancher is started. Additionally, they run every 15 hours when caches are synced. In scaled setups these scheduled runs come with huge performance costs because every handler is being run on every applicable object. However, this scheduled execution of handlers can be disabled using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help.
+
+The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers.
```
cbron, line 40: You should specify whether adding a type to the list enables those controllers to run, or disables them.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Keeping Kubernetes Versions Up to Date
+Similar to Rancher versions, it is advisable to keep your kubernetes cluster up to date. This will ensure that your cluster contains any available performance enhancements or bug fixes.
+
+### Optimizing ETCD
+The two main bottlenecks to [ETCD performance](https://etcd.io/docs/v3.4/op-guide/performance/) are disk speed and network speed. Optimization to either should improve performance. For information regarding ETCD performance see [Slow etcd performance (performance testing and optimization)](https://www.suse.com/support/kb/doc/?id=000020100). Information on disks can also be found [in our docs](https://docs.Ranchermanager.Rancher.io/v2.5/pages-for-subheaders/installation-requirements#disks).
+
+Theoretically, the more nodes in an ETCD cluster the slower it will be due to replication requirements [source](https://etcd.io/docs/v3.3/faq). This may be counter-intuitive to common scaling approaches. It can also be inferred that ETCD performance will be inversely affected by distance between nodes as that will slow down network communication.
```
cbron, line 54: Could also link out here: https://docs.ranchermanager.rancher.io/how-to-guides/advanced-user-guides/tune-etcd-for-large-installs

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: The reason why is because these are grouped into optimization within rancher or outside of rancher. I could move the outside of rancher optimizations above.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,55 @@
+---
+title: Tips for Scaling Rancher
+---
+
+This guide aims to introduce the approaches that should be considered to scale Rancher setups, and associated challenges with doing so. As systems grow performance will naturally reduce, but there are steps we can take to minimize the load put on Rancher, as well as optimize Rancher's ability to handle these larger setups.
+
+## General Tips on Optimizing Rancher's Performance
+* It is advisable to keep Rancher up to date with patch releases. Performance improvements and bug fixes are made throughout the life of a minor release. You can review the release notes to help inform your own decisions on whether an upgrade is necessary but we recommend keeping yourself up to date in most cases.
+
+* Performance will be negatively impacted by increased geographic distance between Rancher's infrastructure and a downstream cluster's infrastructure. If a user or organization requires clusters/nodes all over the world or spread across many regions, it is best to use multiple Rancher installations.
```
moio, line 10:
```suggestion
* Performance will be negatively impacted by increased latency between Rancher's infrastructure and a downstream cluster's infrastructure (eg. geographic distance). If a user or organization requires clusters/nodes all over the world or spread across many regions, it is best to use multiple Rancher installations.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,55 @@
...
+
```
moio, line 11:
```suggestion
* Please always try to scale up gradually, monitoring and observing any change in behavior while doing do. It is usually easier to resolve performance problems as soon as they surface, and before other problems confuse symptoms. 
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
moio: @rmweir I think I agree with @cbron - if I read between the lines correctly what we want to do is be prescriptive about what users should and should not do.

So here's a question for you: what can the user do in order to minimize V1 resource use and/or maximize V2 resource use? Is avoiding to use `/g/clusters` all they can do or is there more?

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,55 @@
...
+The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, and when Rancher is started. Additionally, they run every 15 hours when caches are synced. In scaled setups these scheduled runs come with huge performance costs because every handler is being run on every applicable object. However, this scheduled execution of handlers can be disabled using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help.
+
+The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers.
+* `mgmt` refers to management which only runs on one Rancher node
+* `user` will run for every cluster
+* `scaled` runs on every Rancher node, this is not recommended to be set due to critical functionality the scaled handlers are responsible for
+Adding the controller types to the variable will disable that set of controllers from running their handlers as part of cache resyncing.
```
moio, line 44: I'd suggest to add a "TL;DR"-style line for clarity, eg.:

If you notice CPU usage peaks every 15 hours, add the `CATTLE_SYNC_ONLY_CHANGED_OBJECTS` environment variable to your `rancher` deployment with the value `mgmt,user,scaled`.

moio (commented):
Work goes definitely in the right direction FMPOV, thanks for putting this together 👍 .

Just minor nitpicks from my side.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: Okay. Also, I would say it's `mgmt,user` as we do not recommend `scaled`.

rmweir:
@moio addressed feedback, including providing explicit instructions on what users can do in regards to apps v1 and apps v2.

moio (approved):
The only remaining "open" discussion is https://github.com/rancher/rancher-docs/pull/284/files#r1027830990

I do not think it should prevent merging though - having this text in is already a lot better than nothing. It can be tackled later.

btat (approved):

cbron:
Want to re-review one last time, then will merge.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,61 @@
...
+
+TL;DR: If you notice CPU usage peaks every 15 hours, add the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable to your rancher deployment with the value `mgmt,user`.
```
cbron, line 50: `Tl;DR` is a bit too informal for our docs I would say. Instead we can say `In short, if you....`

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
cbron: I'd still like this to be done ^

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,61 @@
...
+* Only add users to clusters and projects when necessary
+* Remove clusters and projects when they are no longer needed
+* Only use custom roles if necessary
+* Use as few rules as possible in custom roles
+* Consider whether adding a role to a user is redundant
+* Consider that using less, but more powerful, clusters may be more efficient
+* Monitoring the rolebindings created from creating new projects for specific purposes versus creating new clusters
+
+### Using Apps V2 Over Apps V1
+There are two App kubernetes resources that Rancher uses: apps.projects.cattle.io and correct this apps.cattle.cattle.io. V1 here refers to apps.projects.cattle.io since they were introduced first and are legacy, while V2 refers to apps.catalog.cattle.io. Apps V2 are preferrable because they live in the downstream cluster while Apps V1 lives in the local cluster. The dashboard uses Apps V2 while the old `/g/clusters` UI uses Apps V1.
```
cbron, line 32:
```suggestion
There are two App kubernetes resources that Rancher uses: the V1 `apps.projects.cattle.io` and V2 `apps.cattle.cattle.io`. V1 Apps were introduced first in the cluster manager UI and is now a legacy feature. V2 apps are found in the cluster explorer, and are preferable because they live in the downstream cluster while V1 lives in the local cluster.
```
@rmweir See "correct this" string in first sentence above.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,61 @@
...
+
+We recommend removing apps that appear in `/g/clusters`, replacing them with apps in the dashboard if necessary and creating any future apps in the dashboard only.
```
cbron, line 34:
```suggestion
We recommend removing apps that appear in Cluster Manager, replacing them with apps in the Cluster Explorer dashboard if necessary. Any new apps should be created in the dashboard only.
```

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,61 @@
...
+
+Without using ACE, all kubeconfig requests first route through Rancher.
+
+### Experimental: Option to Reduce Event Handler Executions
+The bulk of Rancher's logic occurs on event handlers. These event handlers run on an object whenever the object is updated, and when Rancher is started. Additionally, they run every 15 hours when caches are synced. In scaled setups these scheduled runs come with huge performance costs because every handler is being run on every applicable object. However, this scheduled execution of handlers can be disabled using the CATTLE_SYNC_ONLY_CHANGED_OBJECTS environment variable. If resource allocation spikes are seen on an interval of about 15 hours it is possible this setting can help.
+
+The value for the environment variable can be a comma separated list of the following options. The values refer to types of controllers (the structures that contain and run handlers) and their handlers.
+* `mgmt` refers to management which only runs on one Rancher node
```
cbron, line 45: "management" is still confusing, there is no context as to what it means. Management clusters or controllers or what ? I think the format done below make more sense, like `mgmt controllers will run on a single Rancher node`. We should standardize the base of all 3 sentences.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
```diff
@@ -0,0 +1,61 @@
...
+* `user` will run for every cluster
+* `scaled` runs on every Rancher node, this is not recommended to be set due to critical functionality the scaled handlers are responsible for
+Adding the controller types to the variable will disable that set of controllers from running their handlers as part of cache resyncing.
```
cbron, line 48: No period before "Adding", but I would actually move that whole sentence up to the sentence above the bulleted list as it describes how the list works. Those instructions shouldn't be hidden inside a bullet of the list. If I don't read `scaled` I wouldn't understand how the setting works.

cbron (changes requested):

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: I rewrote this also to no longer use V1 and V2. I think they are confusing because the apps are actually in v3 and v1 api groups respectively and the v1/v2 only really makes sense to us. I replaced v1 and v2 with legacy and new instead.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: I also clarified that the Cluster Explorer should be for the target cluster. Since the legacy apps are installed in local technically.

docs/reference-guides/best-practices/rancher-server/tips-for-scaling-rancher.md
rmweir: I rewrote these to follow similar formats.

cbron (approved):
lgtm, we can do the other things in a follow up PR

PR #346, opened 2022-12-13
Updated local authentication link

Brianalmeida:
Line 30 needed a space to render the file path.

@btat


btat:
Thanks for the PR @Brianalmeida. Since https://github.com/rancher/rancher-docs/pull/345 was created earlier, I went ahead with that one instead.

Brianalmeida:
Sounds good; thank you @btat

PR #348, opened 2022-12-15
Updated Docusaurus install instructions in README

martyav:
Updated Docusaurus install instructions in the README to be more explicit about requirements, add links to required tools, and include note on where official Docusaurus instructions aren't applicable to our docs site. 

Also: 
* Updated headers so that the page doesn't leap from level one headers to level three headers, with no level two headers in between.
* Removed *$* command prompts as unnecessary. They also make it slightly more painful to copy and paste code snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The Rancher Docs website is built with [Docusaurus 2](https://docusaurus.io/), a
 
 The Rancher Docs repository already contains a yarn.lock file, which contains the dependencies you need to build the website.
 
-1. If you haven't already, install [Node]https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/getting-started/install).
+1. If you haven't already, install [Node](https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/getting-started/install).
 1. Fork and clone the rancher-docs repository.
 1. Go into your local rancher-docs folder.
 1. Run `yarn` to install Docusaurus and associated dependencies.
--- a/README.md
+++ b/README.md
@@ -1,47 +1,52 @@
 # Website
 
-This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
+The Rancher Docs website is built with [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
 
-### Installation
+## Installation
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,47 +1,52 @@
...
-$ yarn
-```
+Docusaurus requires [Node](https://nodejs.org/en/) and [Yarn](https://yarnpkg.com/).
+
+1. [Install](https://nodejs.org/en/download/) Node.
+1. [Install](https://yarnpkg.com/getting-started/install) Yarn.
+1. Go into your local rancher-docs repo folder.
+1. Run `yarn add docusaurus` to finish installation.
+
+> **NOTE:** Although the official Docusaurus 2 documentation states that you should run `npx create-docusaurus@latest my-website classic`, this isn't necessary. The repo already contains all the files Docusaurus needs to build the Rancher Docs website.
```
btat, line 14: Users should be able to build a local instance of the docs site using what's in the README. They shouldn't need to consult the Docusaurus docs (and specifically the scaffolding/bootstrapping section where that command is found) if their only goal is to run a local instance. Adding the first sentence may just cause confusion.

README.md
```diff
@@ -1,47 +1,52 @@
...
 
 ```
-$ yarn build
+yarn build
 ```
 
 This command generates static content into the `build` directory and can be served using any static contents hosting service.
 
-### Deployment
+## Deployment
```
btat, line 38: We can drop this section completely since deploying the site is not a user concern.

btat (changes requested):

README.md
martyav: I just pushed a new commit addressing the problems in this section

README.md
```diff
@@ -1,51 +1,38 @@
...
-### Installation
+## Installation
 
-```
-$ yarn
-```
+The Rancher Docs repository already contains a yarn.lock file, which contains the dependencies you need to build the website.
 
-### Local Development
+1. If you haven't already, install [Node]https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/getting-started/install).
```
martyav, line 9:
```suggestion
1. If you haven't already, install [Node](https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/getting-started/install).
```

btat (approved):

PR #359, opened 2023-01-05
Remove outdated reference to members tab

martyav:
Addressing #57

I updated the latest page and the one pertaining to 2.6.

Older versions of the page (2.5, and 2.0 - 2.4) were left alone.


btat (changes requested):
The original issue is looking for updated steps to add project members, but the proposed updates are for cluster members, which is a different concept.

martyav:
Unless I'm still lost, 2.7 still seems to have a Members tab. However, you arrive there by clicking a context menu next to the **Create Namespace** button. I updated accordingly.
add-existing-members-add-member-6

martyav:
@btat Just confirmed that 2.6 has identical UI: 
image

btat (approved):

Issue #378, opened 2023-01-20
K8s >= 1.21, hostNetwork: true needs to be set for tcp-services and udp-services to work.


martyav:
Creating an issue based on an internal request.

> The [documentation](https://rancher.com/docs/rke/latest/en/config-options/add-ons/ingress-controllers/#configuring-network-options) outlines that hostNetwork: true is no longer the default for k8s 1.21 and up. However,  this will cause problems if you have tcp-services and/or udp-services configured. The documentation should note that hostNetwork: true needs to be set for tcp-services and udp-services to work.

> 
> Business impact:

> 
> After upgrading to k8s 1.21.x, tcp-services and udp-services stop working.

martyav:
The original ask points to rancher.com/docs/rke

Although we have similar language in https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-resources-setup/load-balancer-and-ingress-controller/ingress-configuration#docusaurus_skipToContent_fallback and the corresponding page for 2.6, it's not exactly the same. In any case, none of these pages have a warning about TCP- or UDP-services not being able to run unless `hostNetwork` is set to `true`.

Related: I'm not sure if the issue is general or confined to RKE. The problem was validated twice, both on versions of RKE:
* https://github.com/rancher/rancher/issues/35943#issuecomment-1055513521
* https://github.com/rancher/rancher/issues/35943#issuecomment-1029045500

martyav:
Related to https://github.com/rancher/docs/issues/3420

btat:
The source code for rancher.com/docs/rke is currently still in the rancher/docs repo, but as we're about to launch the Docusaurus site for RKE relevant portion should be made to https://github.com/rancher/rke1-docs instead.

PR #383, opened 2023-01-23
Kubernetes >= v1.21, hostNetwork needs to be set to true on Ingress for TCP and UDP Services to work

martyav:
Addresses issue #378 
Related to PR [#3420](https://github.com/rancher/docs/pull/3459)

We also need to update a page in the RKE docset. See issue #378 for more context.


martyav:
@btat Both of these pages contain an in-page link to a missing subsection, entitled, "Use as the default backend"

I took a look at the Git history, but even in the earliest version of the page, this section is missing: https://github.com/rancher/rancher-docs/blob/506e1746430e4fad38c210badcbb148b45caa769/docs/how-to-guides/new-user-guides/kubernetes-resources-setup/load-balancer-and-ingress-controller/ingress-configuration.md#use-as-the-default-backend

Do we go hunting for the missing content, or do we remove the link?

btat (approved):

PR #399, opened 2023-02-03
Add port 80 for hosted k8s port requirements between Rancher nodes

reylejano:
When @frank-at-suse and I were troubleshooting a Rancher cluster on a GKE cluster, when the number of replicas of rancher pods were scaled from 1 to 3 then the pods would not start.
Upon looking into the ingress logs, it had an error when trying to communicate to Rancher on port 80 
This PR updates the port requirements table for Hosted Kubernetes clusters and adds port 80 between Rancher nodes located https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-requirements/port-requirements#ports-for-hosted-kubernetes-clusters
cc: @frank-at-suse

### DIFF
--- a/src/components/PortsImportedHosted.js
+++ b/src/components/PortsImportedHosted.js
@@ -13,7 +13,10 @@ const PortsImportedHosted = () => (
     <tbody>
         <tr>
         <td rowspan="3">Rancher Nodes <sup>(1)</sup></td>
-        <td></td>
+        <td rowspan="3" style={{
+            "background-color": "#3497DA",
+            color: "#ffffff"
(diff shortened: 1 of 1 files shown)


btat (changes requested):
The current change results in an extra column in the table and some info ending up on incorrect cells.

pr399-1

Is the desired outcome something like the image below? 

pr399-2

frank-at-suse:
The first image has the correct information - Port 80 for Rancher Nodes

reylejano:
@btat , I updated to fix the alignment, ptal

btat (approved):

Issue #475, opened 2023-03-07
Update versions table for 2.6.11


btat:
An entry for 2.6.11 needs to be added to https://ranchermanager.docs.rancher.com/versions.

btat:
Closed via #476.

PR #476, opened 2023-03-07
Add 2.6.11 entry to versions table

btat:
Fixes https://github.com/rancher/rancher-docs/issues/475

### DIFF
--- a/src/pages/versions.md
+++ b/src/pages/versions.md
@@ -21,10 +21,10 @@ Below are the documentation and release notes for the currently released version
 
 <table>
   <tr>
-    <td><b>v2.6.10</b></td>
+    <td><b>v2.6.11</b></td>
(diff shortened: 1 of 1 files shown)


martyav (approved):
Looks ok.

We can do this in another PR, but under the Current versions tables, 2.5.17 is the only table missing a third column for the support matrix link. I think we should add a cell simply stating "N/A" in that column to make it clear this isn't an oversight and 2.5.17 doesn't have one.

btat:
> Looks ok.
> 
> We can do this in another PR, but under the Current versions tables, 2.5.17 is the only table missing a third column for the support matrix link. I think we should add a cell simply stating "N/A" in that column to make it clear this isn't an oversight and 2.5.17 doesn't have one.

Just had a closer look and removed 2.5 from the Current versions tables since it's EOL (and has an entry there already).

PR #483, opened 2023-03-09
Update field label to "Container Image" to match UI

kourosh7:
Update section 1. Deploying a Workload> step 7 to say "Container Image" as shown in the UI.


kourosh7:
@btat Can this be reviewed please?

btat (approved):
LGTM. Will merge after verifying how far back this change goes and whether 2.6 docs need to be updated too.

PR #485, opened 2023-03-10
installation-requirements.md: point to latest support matrix

moio:
Current link points to 2.6.3, which is confusing coming from the 2.7 version of the docs.

https://ranchermanager.docs.rancher.com/pages-for-subheaders/installation-requirements

points to:

https://www.suse.com/suse-rancher/support-matrix

which redirects to:

https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-6-3/

Perhaps a better solution would be to change the redirect - up to you to close this PR in that case.


martyav:
I can open another PR addressing the link update

Issue #142, opened 2022-09-15
Authorized Cluster Endpoint for RKE2/K3s


ashawka:
An authorized cluster endpoint allows users to connect to the Kubernetes API server of a downstream cluster without having to route their requests through the Rancher authentication proxy.
The authorized cluster endpoint only works on Rancher-launched Kubernetes clusters. In other words, it only works in clusters where Rancher [used RKE](https://docs.ranchermanager.rancher.io/pages-for-subheaders/launch-kubernetes-with-rancher) to provision the cluster.
https://docs.ranchermanager.rancher.io/reference-guides/rancher-manager-architecture/communicating-with-downstream-user-clusters#4-authorized-cluster-endpoint 

**Feedback**:
- docs says that this is only available on RKE. 
- It is also available for RKE2, while provisioning RKE2 cluster, you can find that option under networking section, which is disabled by default. 
<img width="398" alt="ACE-RKE2" src="https://user-images.githubusercontent.com/34711662/190350969-e57f775f-7449-48fe-9c36-29868f49885e.png">

catherineluse:
I think it should be emphasized what the defaults are because the setting can't be changed after the cluster is created. While ACE is enabled by default for RKE1 and K3s, it is disabled by default for RKE2.

esen-transporeon:
The [guide on how to directly access a cluster ](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/manage-clusters/access-clusters/use-kubectl-and-kubeconfig) was also missing the configuration changes which were needed on rke2, so I created a pull request with a small hint pointing at the right documentation.

PR #496, opened 2023-03-17
Hint for authorized cluster endpoints with rke2 and k3s

esen-transporeon:
Hello,

after it took me a lot of time to find out why my authorized cluster endpoints are not working with rke2 I think it would help other people to create a small hint to the document which states that there needs to be manual changes on rke2 and k3s to support authorized endpoints.

I feel like the guide with the manual changes is a hard to find since it doesn't show up when you search for "rancher authorized cluster endpoint support rke2" on google. Instead [this page](https://ranchermanager.docs.rancher.com/v2.5/how-to-guides/advanced-user-guides/manage-clusters/access-clusters/authorized-cluster-endpoint) shows up. So I think having here a hint would definitely help.

[rancher#142](https://github.com/rancher/rancher-docs/issues/142) is already working on improving the documentation with authorized cluster endpoints and rke2.


btat (changes requested):
I think this hint should go on the 2.6 version of the `user-kubectl-and-kubeconfig` page since the `#authorized-cluster-endpoint-support-for-rke2-and-k3s-clusters` section being linked to doesn't exist on the 2.5 version of the `register-existing-clusters` page being linked to.

btat (approved):

PR #497, opened 2023-03-17
Typo fix for link

martyav:
Typo in in-page link on the Rancher Security page


btat (changes requested):
The same link is broken on other versions. Could you apply the fix to those as well.

btat (approved):

PR #499, opened 2023-03-17
Improve quick start guides via helm

kshtsk:


docs/getting-started/quick-start-guides/deploy-rancher-manager/helm-cli.md
```diff
@@ -15,14 +15,14 @@ The full installation requirements are [here](../../../pages-for-subheaders/inst
 
 ## Install K3s on Linux
 
+Rancher needs to be installed on a supported Kubernetes version. To specify the K3s version, use the INSTALL_K3S_VERSION environment variable when running the K3s installation script. Refer to the [support maintenance terms](https://rancher.com/support-maintenance-terms/).
+
 Install a K3s cluster by running this command on the Linux machine:
 
 ```
-curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="***" sh -s - server --cluster-init
+curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="v1.24.11+k3s1" sh -s - server --cluster-init
```
pgonin, line 23: the K3S version is left blank to be generic, as the version to use might change depending on various requirements

kshtsk: this is right, but it makes hard to figure out which format should it be

docs/getting-started/quick-start-guides/deploy-rancher-manager/helm-cli.md
```diff
@@ -38,6 +38,8 @@ The kubeconfig file is important for accessing the Kubernetes cluster. Copy the
 scp root@<IP_OF_LINUX_MACHINE>:/etc/rancher/k3s/k3s.yaml ~/.kube/config
 ```
 
+In some cases it may need to make sure that your shell has the environment variable `KUBECONFIG=~/.kube/config` defined, for instance, it can be exported in your profile or rc files.
```
btat, line 41: The command suggested is setting the environment variable to the default location. Did you have any example cases (multi-cluster access?) in mind so users know why this would be done? I can suggest updated phrasing when this info is added.

Also, KUBECONFIG is not specific to Linux so it shouldn't be inside the Linux tab.

docs/getting-started/quick-start-guides/deploy-rancher-manager/helm-cli.md
```diff
@@ -15,14 +15,14 @@ The full installation requirements are [here](../../../pages-for-subheaders/inst
 
 ## Install K3s on Linux
 
+Rancher needs to be installed on a supported Kubernetes version. To specify the K3s version, use the INSTALL_K3S_VERSION environment variable when running the K3s installation script. Refer to the [support maintenance terms](https://rancher.com/support-maintenance-terms/).
```
btat, line 18:
```suggestion
Rancher needs to be installed on a supported Kubernetes version. To specify the K3s version, use the INSTALL_K3S_VERSION (e.g., `INSTALL_K3S_VERSION="v1.24.10+k3s1"`) environment variable when running the K3s installation script. Refer to the [support maintenance terms](https://rancher.com/support-maintenance-terms/).
```

Using 1.24.10 instead of proposed 1.24.11 since 1.24.10 is what https://www.suse.com/suse-k3s/support-matrix/all-supported-versions/k3s-v1-24/ currently lists.

btat (changes requested):

docs/getting-started/quick-start-guides/deploy-rancher-manager/helm-cli.md
kshtsk: Here is the basic scenario when you have a node and want to use kubectl on the same machine and not using the root:
- spin up a vm with ssh access via opensuse (any opensuse image, for SLE the user is 'sees', for Ubuntu distroes it is ubuntu) user
- install k3s
- copy /etc/rancher/k3s/k3s.yaml to opensuse user home
- try to run kubectl under opensuse user
- since /etc/rancher/k3s/k3s.yaml present and cannot be read because of permission, kubectl and helm will fail

btat: > These instructions assume you have a Linux virtual machine that you will communicate with from your local workstation. Rancher will be installed on the Linux machine. You will need to retrieve the IP address of that machine so that you can access Rancher from your local workstation.

The above block is from the start of this page. which is a different setup than your example. Looking at the [K3s docs](https://docs.k3s.io/cluster-access#accessing-the-cluster-from-outside-with-kubectl), it doesn't seem like the suggested change is needed either.

kshtsk: This doc is not about k3s, this is doc about quick start for rancher, which means in some cases zero knowledge about k3s. So your suggested user instead of quick start, for example trying on his own rancher should invest in prior learning of k3s before they are ready to continue with rancher?

btat: > This doc is not about k3s, this is doc about quick start for rancher, 

That's not entirely correct as our quick starts are mostly in the form of "Rancher + X" (e.g. Rancher AWS Quick Start, Rancher GCP Quick Start, or Rancher + Helm in this case), but how we classify the quick starts is not really the point of this thread.

> which means in some cases zero knowledge about k3s. So your suggested user instead of quick start, for example trying on his own rancher should invest in prior learning of k3s before they are ready to continue with rancher?

I'm not suggesting any prior K3s knowledge is needed as the required information is already on the page. I'm actually suggesting the proposed change shouldn't be added because it's to address something for a setup different than the one on this page. The link to the K3s docs was to show that the scenario on this page does not require setting the environment variable. This link was intended as a reference for this thread and suggesting users need to look at the K3s docs.

btat (approved):

PR #522, opened 2023-04-04
Adding GPU limit/reservation to set-container-default-resource-limits.md

ashawka:
From Rancher UI > Project > Create > Container Default Resource limit  I can see already that NVIDIA GPU limit/reservation is available, although from the resource type section, there is not mention to the GPU


btat:
Added missing description and applied update to 2.6 as well.

Ref: https://github.com/rancher/dashboard/issues/5005.

btat (approved):

PR #530, opened 2023-04-11
Update note about upstream CIS K8s v1.25 benchmark

macedogm:
Update note about upstream CIS K8s v1.25 benchmark availability in Rancher and the hardening guides.

When the hardening guides were updated, the official benchmark for K8s v1.25 was in draft mode by CIS. Now it's already available (it was released 2-3 week ago).


LucasSaintarbor (approved):
LGTM 👍

PR #543, opened 2023-04-17
#207 Special note on rhel/centos provisioning

### LINKED ISSUE #207: Provisioning redhat/centos  by preventing "yum install curl"
Information about the Cluster

Kubernetes version: v1.23.10
Cluster Type (Local/Downstream): downstream
If downstream, what type of cluster? (Custom/Imported or specify provider for Hosted/Infrastructure Provider): RKE VMware vSphere but should effect any node-driver cluster.
Cluster installed by using nodetemplate

**Issue**

When provisioning a new downstream cluster with RHEL nodes, the provision phase doesn't work, it fails on the installation of curl even if curl is already installed on the node.

**To Reproduce**

Create a new downstream cluster with RHEL nodes where the docker is pre-installed in the image or no repositories are added.

**Result**

the provisioning phase fails with this error message
Error creating machine: RHEL ssh command error: command: sudo -E yum install -y curl err: exit status 1 output: Updating Subscription Management repositories.

**Fix**

In node template under engine option use Docker install as "None" . This will bypass the checking of curl and docker installation

image

Spin up the node after updating this in the node template

martyav:
Fixes #207 

As noted in https://github.com/rancher/rancher/issues/27646#issuecomment-769542772, the default value will work, as well as **none**.

To-do: versioned_docs versions of relevant pages


docs/reference-guides/cluster-configuration/downstream-cluster-configuration/node-template-configuration/amazon-ec2.md
```diff
@@ -47,3 +47,7 @@ If you need to pass an **IAM Instance Profile Name** (not ARN), for example, whe
 ### Engine Options
 
 In the **Engine Options** section of the node template, you can configure the container daemon. You may want to specify the container version or a container image registry mirror.
+
+#### Red Hat Enerprise Linux (RHEL) & CentOS Provisioning
```
martyav, line 51:
```suggestion
#### Red Hat Enterprise Linux (RHEL) & CentOS Provisioning
```

slickwarren (commented):
lgtm

LucasSaintarbor (approved):
@martyav LGTM 👍

btat (approved):

martyav:
Writing up a generic warning is more challenging than I expected, since the node template fields presented to users depend on the node type. What fields are available in the UI aren't the same if they pick EKS, EC2, AKS, Google, etc. and getting through the wizard sometimes requires credentials for the cloud services. The wizard for EC2 doesn't even seem to have this field, and we have a page on Nutanix even though it's not selectable through the node template UI for RKE.

martyav:
I went and checked. The following node template wizards include the **Docker Install URL** field under **Engine Options**:

- Azure
- Vsphere
- Nutanix <-- Not available by default!
- Harvester <-- No docs page

The following node template wizards DON'T include that field:

- Digital Ocean
- Linode <-- No docs page

Amazon EC2 is ambiguous. Instead of having the wizard all on one screen, it's split up into steps, and to continue from step one, you need to add credentials.

docs/reference-guides/cluster-configuration/downstream-cluster-configuration/node-template-configuration/amazon-ec2.md
```diff
@@ -44,6 +45,6 @@ Configure the instances that will be created. Make sure you configure the correc
 
 If you need to pass an **IAM Instance Profile Name** (not ARN), for example, when you want to use a [Kubernetes Cloud Provider](../../../../pages-for-subheaders/set-up-cloud-providers.md), you will need an additional permission in your policy. See [Example IAM policy with PassRole](#example-iam-policy-with-passrole) for an example policy.
 
-### Engine Options
+## Engine Options
```
martyav, line 48:
```suggestion
### Engine Options
```

PR #548, opened 2023-04-19
Updated reference to 2.6 installation requirements in the 2.7 docs

juanbrny:
Replaced this url https://rancher.com/docs/rancher/v2.6/en/installation/requirements/ by the new one for Rancher 2.7: https://ranchermanager.docs.rancher.com/pages-for-subheaders/installation-requirements

PR #555, opened 2023-04-21
Follow-up to #548: convert to relative link

btat:
This is a follow-up to https://github.com/rancher/rancher-docs/pull/548. Since the modified link is an internal link, a relative link should be used to allow for future maintainability (e.g. versioning between pages, domain changes).


martyav (approved):
Confirmed link works

Issue #420, opened 2023-02-10
Search engine results show older versions as top result


gunamata:
**Summary:**

~~Explore approaches to implement redirect docs links from search engine results pages to latest versions~~

~~For example, Clicking on https://ranchermanager.docs.rancher.com/v2.5/pages-for-subheaders/backup-restore-and-disaster-recovery from a search engine results page should automatically take you to https://ranchermanager.docs.rancher.com/pages-for-subheaders/backup-restore-and-disaster-recovery~~

When using search engines to find documentation, sometimes an older version of a doc will be the top result of a search instead of the latest. Behavior varies with the search term used.

E.g. (Mar 2, 2023) Searching `rancher backup restore` gives https://ranchermanager.docs.rancher.com/v2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-rancher as the top result even though there is https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-rancher available.

btat:
Based on discussion the proposed approach is to test using [single page metadata](https://docusaurus.io/docs/seo#single-page-metadata) (e.g. `keywords` and `description` in the front matter block) to increase visibility/SEO of the latest pages.

Due to the size of the task it should be broken down into subtasks:
1. Split by version, with the latest (/docs) version being added first. For versionsed docs the keywords/descriptions will replicate what was added to the latest version, but with a version number attached. E.g. If the keyword used for a page on /docs is `install Rancher` then for version 2.6 it'd be `install Rancher 2.6`.
2. Within each version further split based on directories.

martyav:
Here are links to official Google and Bing docs on SEO. I couldn't find anything official from DuckDuckGo.

* https://developers.google.com/search/docs/fundamentals/seo-starter-guide
* https://www.bing.com/webmasters/help/webmaster-guidelines-30fba23a

btat:
More notes:
- A sitemap is generated automatically (using the plugin mentioned in the previous comment). 
  - https://ranchermanager.docs.rancher.com/sitemap.xml
- `rel="canonical"` is automatically added to every page.
  - Docusaurus doesn't provide customization options or the ability to remove adding the annotation to a page. 
  - All versions of a page has the annotation added though. E.g. for /faq/general-faq
    - /docs: `rel="canonical" href="http://ranchermanager.docs.rancher.com/faq/general-faq"`
[...]

PR #604, opened 2023-05-15
#420 Test Docusaurus head metadata override of autogenerated canonical annotations

martyav:
When sites have multiple versions of the same/similar pages, Google uses the `canonical` annotation in the page's `head` to determine which URL to elevate in search rankings. The annotation is accompanied by an URL, which indicates which page to rank higher. For example, if we have two pages, A and B, and we want A to have the higher search ranking, we would set 
`<link rel="canonical" href="httpts://A.com"/>` on A and `<link rel="canonical" href="httpts://A.com"/>` on B as well.

By default, Docusaurus automatically adds the annotation to every page in a docset, but the URL always points to the address of the page it's set on. To go back to our example, this results in Docusaurus setting `<link rel="canonical" href="httpts://A.com"/>` on A and `<link rel="canonical" href="httpts://B.com"/>` on B.

If you have multiple versions of the same page, Docusaurus will create competing canonical pages, impairing search results. 

In our case, this means that `/versioned_docs` pages erroneously point to themselves as canonical, and search results will often give users older pages instead of pages pertaining to the latest version of Rancher.

Thankfully, Docusaurus allows us to [override ](https://docusaurus.io/docs/2.3.1/markdown-features/head-metadata) this behavior by editing each page's `head`. This is distinct from the front matter we can add to our Markdown files, as this metadata is contained within standard HTML tags.

Here's a screenshot of the page source on a preview build of the v2.5 page we edited for the PR:

image

As you can see, it points to the URL for the v2.7 page in `/docs`.

Hopefully, this will help address problems with users finding older pages within search results.

### DIFF
--- a/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
+++ b/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
@@ -3,7 +3,7 @@ title: Rancher Helm Chart Options
 ---
 
 <head>
-    <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options"/>
+  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options"/>
 </head>
 
 This page is a configuration reference for the Rancher Helm chart.
--- a/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
+++ b/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
@@ -2,6 +2,10 @@
 title: Rancher Helm Chart Options
 ---
 
+<head>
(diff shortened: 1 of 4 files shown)


docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
```diff
@@ -2,6 +2,10 @@
 title: Rancher Helm Chart Options
 ---
 
+<head>
+    <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options"/>
```
btat, line 6:
```suggestion
  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/installation-references/helm-chart-options"/>
```

btat (approved):
LGTM. Small nit, but suggest using 2 spaces for indentation to follow the Docusaurus example and just for consistency with our Markdown.

PR #605, opened 2023-05-15
Add keywords to metadata to test SEO impact

btat:
Related to #420 

When using certain queries/keywords to a perform a search, the resulting top result is not as expected, as outlined in [this table](https://github.com/rancher/rancher-docs/issues/420#issuecomment-1452454114).

This PR is to test whether the addition of keywords in the metadata can help improve search results so that
- the correct page ranks higher
- the latest version of a page ranks higher (if no version is specified)
- produce consistent results across search engines

### DIFF
--- a/versioned_docs/version-2.0-2.4/pages-for-subheaders/backup-restore-and-disaster-recovery.md
+++ b/versioned_docs/version-2.0-2.4/pages-for-subheaders/backup-restore-and-disaster-recovery.md
@@ -1,5 +1,6 @@
 ---
 title: Backups and Disaster Recovery
+keywords: [rancher v2.0-v2.4 backup restore, rancher v2.0-v2.4 backup and restore, backup restore rancher v2.0-v2.4, backup and restore rancher v2.0-v2.4]
 ---
 
 This section is devoted to protecting your data in a disaster scenario.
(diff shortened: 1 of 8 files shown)


versioned_docs/version-2.0-2.4/pages-for-subheaders/backup-restore-and-disaster-recovery.md
```diff
@@ -1,5 +1,6 @@
 ---
 title: Backups and Disaster Recovery
+keywords: [rancher v2.0-v2.4 backup restore, rancher v2.0-v2.4 backup and restore, backup restore rancher v2.0-v2.4, backup and restore rancher v2.0-v2.4]
```
martyav, line 3: Users are probably more likely to search for their version number rather than a range.

martyav: Including the same terms with each individual version number four times over would get very verbose, though. I think using the range is fine for this PR.

martyav (approved):
I left some comments about using a range for pages pertaining to 2.0 through 2.4. While users are more likely to search for their exact version number, including the same terms with different version numbers 4 times over on the same page would 1) be very verbose and potentially distracting 2) might mess with search results (because the same keywords are appearing repeatedly, they might weigh the page more heavily -- or get interpreted as keyword spam and weighed more lightly).

PR #611, opened 2023-05-18
Bump actions/checkout. v2 uses deprecated Node 12

btat:
Seeing this message `Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.` in a recent [failed build](https://github.com/rancher/rancher-docs/actions/runs/5018572437).

Bumping to v3 to see if it resolves this.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ jobs:
     name: Deploy to GitHub Pages
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


martyav (approved):
I looked at our action logs and confirmed that the run associated with this PR was the only one not to throw a deprecation warning

PR #626, opened 2023-05-24
#420 Adding canonical refs to ./getting-started part 7/10

martyav:
See https://github.com/rancher/rancher-docs/pull/619 for more info


btat (approved):
First 4 commits are already covered in #625.

Issue #663, opened 2023-05-18
[BUG] Hyperlink broken for Private AKS cluster creation details


salasberryfin:
**Rancher Server Setup**
- Rancher version: [2.7.5-rc1](https://github.com/rancher/rancher/releases/tag/v2.7.5-rc1)
- Installation option: **Helm**
   - If Helm Chart, Kubernetes Cluster and version: **AKS with Kubernetes v1.25.5**

**Information about the Cluster**
- Kubernetes version: **v1.26.0**
- Cluster Type: **AKS cluster**

**User Information**
- What is the role of the user logged in? **Admin**

**Describe the bug**
This issue was found during [validation of version 2.7.5-rc1](https://github.com/rancher/rancher/issues/41541).

When provisioning a private AKS cluster (selecting the 'Enable Private Cluster' option), a pop-up appears with the following message:

```text
AKS private clusters is an advanced use case which requires Rancher to be running on the same NAT as the AKS 
nodes. Before enabling private clusters ensure you have read through the AKS cluster provisioning docs for 
instructions on configuring AKS so nodes may communicate with Rancher.
```

Including the link [AKS cluster provisioning docs](https://rancher.com/docs/rancher/v2.7/en/cluster-provisioning/hosted-kubernetes-clusters/aks/), which returns a 404 error.

**To Reproduce**
- Using Rancher UI, provision a new AKS cluster
- Configure the cluster to be private: Networking Options -> Show advanced options -> Enable Private Cluster

**Result**
Link to documentation [AKS cluster provisioning docs](https://rancher.com/docs/rancher/v2.7/en/cluster-provisioning/hosted-kubernetes-clusters/aks/) returns `404 - Not found`

**Expected Result**
Hyperlink should be updated for Private AKS cluster creation details

**Screenshots**
![image](https://github.com/rancher/aks-operator/assets/10220061/8571b733-bccb-4607-9c3c-9321de32af3a)

**Additional context**
This issue has also been identified for other providers:
- rancher/rancher-docs#662

btat:
Redirects have been added.

@nwmac @gaktive we now have a versioned link for all docs so hopefully that makes things easier for the UI team.

Issue #662, opened 2023-05-18
[BUG] Hyperlink broken for Private GKE cluster creation details


cpinjani:
**Rancher Server Setup**
Rancher version: 2.7.5-rc1
Installation option: Helm Chart
If Helm Chart, Kubernetes Cluster and version: GKE with Kubernetes 1.26.3-gke.1000

**Information about the Cluster**
Kubernetes version: GKE with Kubernetes 1.26.3-gke.1000
Cluster Type (Downstream): Hosted = GKE

**User Information**
What is the role of the user logged in? - Cluster Owner

**Describe the bug**
Creation of node pools with same name must be handled gracefully. While provisioning new GKE cluster, it starts provisioning and fails. While editing existing cluster, it starts updating and fails. Both scenarios should be restricted for user to perform.

**To Reproduce**
1. Add valid GCP cloud credentials
2. Provision downstream GKE cluster with private option, click on link on the page [GKE cluster provisioning docs](https://rancher.com/docs/rancher/v2.7/en/cluster-provisioning/hosted-kubernetes-clusters/gke/)
It navigates to https://rancher.com/docs/rancher/v2.7/en/cluster-provisioning/hosted-kubernetes-clusters/gke/

**Result**
Hyperlink broken for Private GKE cluster creation details

**Expected Result**
Hyperlink should be updated for Private GKE cluster creation details

**Screenshot**
![image](https://github.com/rancher/gke-operator/assets/73099870/d31db399-2719-47de-bbba-da1531efc6db)

gaktive:
Some connection to https://github.com/rancher/rancher-docs/issues/663

PR #665, opened 2023-06-05
Add 2.7 to versioned docs

btat:
Related: #662, #663
Note: `versioned_docs/version-2.7` generated when 05f194a3f35ec7fe5db1ed0a7fa5902f132d7587 the most recent commit in /docs.

This PR adds a versioned_docs/ entry for 2.7 so that there are versioned URLs for 2.7 pages, which is used by things like the UI. 

With this setup, `/docs` and the latest version in `versioned_docs` are the same (`versioned_docs/version-2.7` in this case) but will be appear to users labeled as `Latest` and `v2.7` respectively. For example, if v2.8 is introduced, a `versioned_docs/version-2.8` folder will be created and that will be a replica of `/docs` .


martyav:
The official Docusaurus docs suggest updating the config so that we don't have to maintain 2 folders. I haven't tested that out myself on our docset so I don't know if there would be any problems implementing that:

image

From https://docusaurus.io/docs/versioning#configuring-versioning-behavior

btat:
The example quoted is the setup we currently have, except that we don't set `current.path`.

Using the `/faq/general-faq` page as an example:
- Our current setup gives us https://ranchermanager.docs.rancher.com/faq/general-faq
- Based on feedback, we need both https://ranchermanager.docs.rancher.com/faq/general-faq and https://ranchermanager.docs.rancher.com/v2.7/faq/general-faq
  - A versioned link for things like the UI which need a static link
  - The unversioned link for things like rancher.com or marketing where we want users to access the latest page 

`path` doesn't take an array so we can't set it to both `/` and `/v2.7` which is why another folder in `versioned_docs` is needed. [Rancher Desktop](https://github.com/rancher-sandbox/docs.rancherdesktop.io/) has been using this model for a while now.

btat:
Build is currently failing with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory` as shown in https://github.com/rancher/rancher-docs/actions/runs/5181336107/jobs/9336713418?pr=665.

After some local testing, using the combination #667 and #668 allowed me to build without running into memory issues. Those should be merged before this one.

martyav:
A more radical approach: Putting all content for latest in /shared-files, and importing it to the respective pages in /docs and /version_2.7. Since that would take more work than simply copying the files over to a new folder, we should probably leave off on it until 2.8 hits. However, it would give us a single source of truth, so that pages can't get out of sync, and it would dramatically reduce the size of the pages stored

btat:
Rebased and applied 9f65c72d..ae741d3b to version-2.7 docs.

martyav (approved):

PR #685, opened 2023-06-13
#669 New flag for GKE ingress

### LINKED ISSUE #669: We have a new flag change while using GKE ingress, update the rancher installation on the GKE clusters
**Summary:**
Please add additional info in the following page:
https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/rancher-on-gke 

To successfully install Rancher on GKE clusters using a GKE ingress, it is necessary to include the following flag in the installation command: --set service.type: NodePort

Reference issue: https://github.com/rancher/rancher/issues/16061

martyav:
Fixes #669 

I know this needs to be mirrored to versioned_docs/2.7, I just want to verify that the edits look good.

### DIFF
--- a/docs/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/rancher-on-gke.md
+++ b/docs/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/rancher-on-gke.md
@@ -153,7 +153,6 @@ helm upgrade --install \
   ingress-nginx ingress-nginx/ingress-nginx \
   --namespace ingress-nginx \
   --set controller.service.type=LoadBalancer \
-  --set service.type: NodePort \
   --version 4.0.18 \
   --create-namespace 
 ```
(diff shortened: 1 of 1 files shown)


martyav:
@btat Sorry about that, moved too fast. I'll update accordingly

martyav:
...and now that you've pointed it out, I think the release note for this also uses `:` instead of `=`.

docs/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/rancher-on-gke.md
```diff
@@ -188,12 +188,16 @@ Next, install the Rancher Helm chart by following the instructions on [this page
...
-When installing Rancher on top of this setup, you will also need to pass the value below into the Rancher Helm install command in order to set the name of the ingress controller to be used with Rancher's ingress resource:
+When installing Rancher on top of this setup, you will also need to set the name of the ingress controller to be used with Rancher's ingress resource:
 
 ```
 --set ingress.ingressClassName=nginx
 ```
 
 Refer [here for the Helm install command](../../../pages-for-subheaders/install-upgrade-on-a-kubernetes-cluster.md#5-install-rancher-with-helm-and-your-chosen-certificate-option) for your chosen certificate option.
+
+In Rancher v2.7.5, if you intend to use the default GKE ingress on your cluster, you need to set the following flag:
```
btat, line 199: We need to include that this is only relevant  "when using the default GKE ingress in a cluster not using VPC-native cluster mode (ie `--no-enable-ip-alias` is set) it is incompatible with the ClusterIP type" as per this [comment](https://github.com/rancher/rancher/issues/16061#issuecomment-1579352835).

btat (changes requested):

btat (approved):

PR #696, opened 2023-06-22
Update nginx-load-balancer.md https upstream on wrong block

jaynakus:


brandond (approved):
This moves the `upstream rancher_servers_https` block into the `http` block where it is used, instead placing it in the `stream` block which uses the previously-defined `upstream rancher_servers_http`.

LGTM!

brandond:
@jaynakus thanks for the contribution, but in the future it might be helpful if you link to an issue or at least put some commentary in the PR body describing the nature of your change. This is helpful for maintainers of the project that might not be familiar with nginx configuration syntax.

PR #697, opened 2023-06-22
Update second server node setup description in HA cluster

alex1704:
Current description lacks steps when you have to obtain and apply master node token for second server.

### DIFF
--- a/docs/how-to-guides/new-user-guides/kubernetes-cluster-setup/k3s-for-rancher.md
+++ b/docs/how-to-guides/new-user-guides/kubernetes-cluster-setup/k3s-for-rancher.md
@@ -45,7 +45,17 @@ When running the command to start the K3s Kubernetes API server, you will pass i
 
   :::
 
-1. Repeat the same command on your second K3s server node.
+1. Get main server node token:
+  ```
+  cat /var/lib/rancher/k3s/server/node-token
+  ```
+
+1. Run command on your second K3s server node:
+  ```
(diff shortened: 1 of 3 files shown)


docs/how-to-guides/new-user-guides/kubernetes-cluster-setup/k3s-for-rancher.md
```diff
@@ -45,7 +45,17 @@ When running the command to start the K3s Kubernetes API server, you will pass i
 
   :::
 
-1. Repeat the same command on your second K3s server node.
+1. Get main server node token:
+  ```
+  cat /var/lib/rancher/k3s/server/node-token
```
btat, line 50:
```suggestion
  cat /var/lib/rancher/k3s/server/token
```

Looking at the [K3s docs](https://docs.k3s.io/datastore/ha#3-optional-join-additional-server-nodes), it shows `token` instead of `node-token`. @alex1704 are you seeing `node-token` in your setup?

btat (changes requested):
Note to self/docs team, changes will need to be applied to other version after review is done.

docs/how-to-guides/new-user-guides/kubernetes-cluster-setup/k3s-for-rancher.md
alex1704: @btat I see both and they are equal.

btat (approved):

PR #715, opened 2023-06-30
#420 SEO Improvements: Helm charts in Rancher directory. Part 1/1

martyav:
Another small one

### DIFF
--- a/docs/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps.md
+++ b/docs/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps.md
@@ -2,6 +2,10 @@
 title: Creating Apps
 ---
 
+<head> 
+  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps"/>
+</head>
+
 Rancher's App Marketplace is based on Helm Repositories and Helm Charts. You can add HTTP based standard Helm Repositories as well as any Git Repository which contains charts.
 
 :::tip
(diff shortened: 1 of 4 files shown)


btat:
For 2.0-2.4, there's https://ranchermanager.docs.rancher.com/v2.0-v2.4/how-to-guides/new-user-guides/helm-charts-in-rancher/creating-apps, which talks about `Creating Catalog Apps` and the content on the page is largely similar to the ones in this PR. The Apps Marketplace was the next iteration/replacement to the Catalog service. How have we been handling canonical links for these types of cases? Have we been using the replacement feature's page as the canonical or only using the exact same page of a newer version as the canonical?

martyav:
My workflow has been to go through the files in /docs, then find the files in /versioned_docs whose title front matter matches  the title of the /docs file. I did this on the theory that older pages with the same title are most likely to harm the latest pages' search ranking. This workflow is able to find pages whose file path has changed across versions, but it doesn't account for  pages with different titles/similar content. Since I start from /docs, pages in versioned_docs that don't cleanly match with pages in /docs might get missed. Sometimes I get a hint of these cases while performing a search, but not always. I think we can account for more of these cases by doing another sweep targeting versioned_docs with missing canonical tags. But since we have a known case here, I can add those files to this PR.

btat:
> @btat My workflow has been to go through the files in /docs, then find the files in /versioned_docs whose title front matter matches the title of the /docs file. I did this on the theory that older pages with the same title are most likely to harm the latest pages' search ranking. This workflow is able to find pages whose file path has changed across versions, but it doesn't account for pages with different titles/similar content. Since I start from /docs, pages in /versioned_docs that don't cleanly match with pages in /docs might get missed. Sometimes I get a hint of these cases while performing a search, but not always. I think we can account for more of these cases by doing another sweep targeting versioned_docs with missing canonical tags. But since we have a known case here, I can add those files to this PR.

Thanks @martyav , the current workflow you described sounds good. I just wanted to confirm since I can't recall if we discussed the approach in the past. For this particular case, I just happened to be aware of that there's an existing page.

martyav:
@btat added 2.0-2.4 version of page

btat (approved):

PR #721, opened 2023-07-03
toggled dark theme switch

martyav:


btat:
This was one of the topics that we talked about when we did the migration to Docusaurus (as well as when we did a Docusaurus PoC with Rancher Desktop docs), but I can't recall the reason behind the decision. Let's wait till Guna is back to see if he remembers why before we merge.

martyav:
Oops -- this was meant to target my fork so I could more easily access a preview through the Netlify integration I set up. I was having trouble building locally. Nevertheless, we should have this conversation again as a user opened an issue about it recently and other docs sites, such as [Mirantis](https://docs.mirantis.com/container-cloud/latest/overview.html), have it set up.

martyav:
Fwiw, both my local build and the Netlify preview failed. Hopefully it's caused by something that wouldn't effect production if we merged this...

PR #759, opened 2023-08-08
#755 Update Note for CIS Chart Installation on hardened cluster

### LINKED ISSUE #755: Update Note for CIS Chart Installation on hardened cluster
**Summary:**
The note in the Rancher documentation [here](https://ranchermanager.docs.rancher.com/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark) needs to be updated to explicitly specify that it pertains to PSP hardened clusters as it is not relevant to PSA hardened clusters (K8s version >= 1.25). It only applies to PSP hardened clusters (K8s version < 1.25):
`CIS Benchmark 4.0.0 and above have PSPs disabled by default. To install CIS Benchmark on a hardened cluster, set global.psp.enabled to true in the values before installing the chart.`

cc @btat

martyav:
Fixes #755 

Rather than update the note, I opted for tabs, since later Kubernetes versions have an alternate path to applying the benchmarks. I made the instructions for the earlier Kubernetes version first since that's what's already listed on the page, but we can switch the order if you think we should emphasize the latest version.

I also updated the PSA config template page to state that assigning a template is how you apply the CIS benchmark

### DIFF
--- a/docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
+++ b/docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
@@ -25,6 +25,3 @@ CIS Benchmark 4.0.0 and above have PSPs disabled by default. To install CIS Benc
 If you have a cluster running Kubernetes v1.25 or later, you can set the Rancher CIS benchmark by [assigning](../../new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates.md) the `rancher-restricted` [PSA](../../new-user-guides/authentication-permissions-and-global-configuration/pod-security-standards.md) configuration template to the cluster and selecting a CIS profile.  
 </TabItem>
 </Tabs>
-
-[dgf](../open-ports-with-firewalld.md)
-
--- a/docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
+++ b/docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
@@ -2,6 +2,10 @@
 title: Install Rancher CIS Benchmark
 ---
 
+<Tabs>
+<TabItem value="Kubernetes v1.24 and earlier">
+The following instructions apply to PSP-hardened clusters, which were available in Kubernetes v1.24 and earlier.
+
 1. In the upper left corner, click **☰ > Cluster Management**.
 1. On the **Clusters** page, go to the cluster where you want to install CIS Benchmark and click **Explore**.
 1. In the left navigation bar, click **Apps > Charts**.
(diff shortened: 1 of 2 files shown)


docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
```diff
@@ -15,3 +19,9 @@ title: Install Rancher CIS Benchmark
 CIS Benchmark 4.0.0 and above have PSPs disabled by default. To install CIS Benchmark on a hardened cluster, set `global.psp.enabled` to `true` in the values before installing the chart.
 
 :::
+
+</TabItem>
+<TabItem value = "Kubernetes v1.25+">
+If you have a cluster running Kubernetes v1.25 or later, you can set the Rancher CIS benchmark by [assigning](../../new-user-guides/authentication-permissions-and-global-configuration/psa-config-templates.md) the `rancher-restricted` [PSA](../../new-user-guides/authentication-permissions-and-global-configuration/pod-security-standards.md) configuration template to the cluster and selecting a CIS profile.
```
btat, line 25: This page is about installing the CIS Benchmark Helm Chart (and the note is about modifying the chart), which is a different topic than the content being added to this tab. If we're addressing #755 using tabs, this tab should have the same instructions minus the note.

Also, we should have the more recent tab come first.

btat (changes requested):

docs/how-to-guides/advanced-user-guides/cis-scan-guides/install-rancher-cis-benchmark.md
martyav: But if the instructions only apply to PSP-hardened clusters, how could they apply to clusters running Kubernetes v1.25 and later? PSPs were removed in v1.25. @Priyashetty17

btat: > But if the instructions only apply to PSP-hardened clusters, how could they apply to clusters running Kubernetes v1.25 and later? PSPs were removed in v1.25.
> 
> From #755:
> 
> > it is not relevant to PSA hardened clusters (K8s version >= 1.25)
> 
> @Priyashetty17

Hence the `minus the note` bit in my comment above. #755 is referring to the note only, not the entirety of the instructions. The note calls out setting a Helm value to turn on PSPs for the chart and the PSP portion is the only thing not applicable to K8s >= 1.25 on this page. The rest of the step-by-step instructions to install the chart aren't affected by whether its PSP or PSA. The new tab for v1.25+ removes how to install the chart, which is the purpose of the page.

martyav:
Closing as the branch doesn't adequately address the issue. Taking a note of it for the next PR.

PR #761, opened 2023-08-08
#686 clarify when values should be raw or manually base64 encoded

### LINKED ISSUE #686: Clarify when values should be raw or manually base64 encoded
**Summary:**
Currently the example for [S3 Credential Secrets](https://ranchermanager.docs.rancher.com/reference-guides/backup-restore-configuration/examples#example-credential-secret-for-storing-backups-in-s3) for Rancher Backups has the text "Enter your base64 encoded..." in it, which is confusing/unclear. This is only accurate for if the user is manually creating the Secret via kubectl or similar tools, not through the UI. The docs could use some clarification on when the user should manually base64 encode values and when they should use raw values.

I'm not sure if there are other areas in the docs where similar verbiage is used but some kind of note for when values entered via the UI should be raw vs base64 encoded would be helpful!

martyav:
Fixes #686.

This addresses one instance, but not the broader issue. I'll make a note of it in our internal docs to clarify when values should be entered as-is or be encoded.


docs/reference-guides/backup-restore-configuration/examples.md
```diff
@@ -244,6 +244,8 @@ spec:
 
 ## Example Credential Secret for Storing Backups in S3
 
+The `accessKey` and `secretKey` must be base64 encoded values. If you acquired your `accessKey` or `secretKey` through the Rancher UI, enter the value exactly as it appears in the UI. If you acquired either key through a commandline tool such as kubectl, you'll need to manually encode the key in base64.
```
btat, line 247: I don't fully agree with the concern brought up by #686 for this particular example since we're representing the secret using YAML here and the data field is base64 encoded per https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#secret-v1-core. Even if consider secret creation via "Edit as YAML" as using the UI, the input for `data` still needs to be base64 encoded. 

In any case, the current PR doesn't really address the concern brought up in #686 though. `The accessKey and secretKey must be base64 encoded values` is what #686 is suggesting isn't always true. The access key and secret key are referring to AWS keys, which you wouldn't obtain though the Rancher UI nor kubectl. I believe #686 is suggesting that we have something along the lines of:

If creating the secret using the UI (the form), enter the accessKey and secretKey as raw/plaintext.
If creating the secret using other methods, enter the accessKey and secretKey in their base64 encoded form.

But I think the simplest solution is to just remove the section from this page. The [backup-restore-configuration/backup-configuration](https://ranchermanager.docs.rancher.com/reference-guides/backup-restore-configuration/backup-configuration#example-credentialsecret) page already has this same sample YAML block, but with an explicit kubectl sample command on using it. We can amend that portion with a link to creating a secret though the UI as well. The section from this page is just a duplicate.

btat (changes requested):

docs/reference-guides/backup-restore-configuration/examples.md
martyav: Thanks for the thorough review. We should definitely reduce duplication across pages as that quickly adds up

docs/reference-guides/backup-restore-configuration/backup-configuration.md
```diff
@@ -79,9 +85,11 @@ The S3 storage location contains the following configuration fields:
...
 | ---------------- | ---------------- | ------------ |
-| `credentialSecretName` |  If you need to use the AWS Access keys Secret keys to access s3 bucket, create a secret with your credentials with keys and the directives `accessKey` and `secretKey`. It can be in any namespace as long as you provide that namespace in `credentialSecretNamespace`. An example secret is [here.](#example-credentialsecret) This directive is unnecessary if the nodes running your operator are in EC2 and set up with IAM permissions that allow them to access S3, as described in [this section.](#iam-permissions-for-ec2-nodes-to-access-s3)  |  |
+| `credentialSecretName` |  If you need an AWS access key or secret key to access an S3 bucket, [create a secret](../../how-to-guides/new-user-guides/kubernetes-resources-setup/secrets.md) using your credentials, with keys and directives named `accessKey` and `secretKey`. The secret can be in any namespace. An example secret is [here](#example-credentialsecret). This directive is unnecessary if the nodes running your operator are in EC2 and assigned [IAM permissions to access S3](#iam-permissions-for-ec2-nodes-to-access-s3). |  |
```
btat, line 92: This section is for the YAML fields, but we don't mention YAML at all on secrets.md so linking to it might be confusing.

btat (dismissed):

PR #771, opened 2023-08-11
#420 SEO Improvements: Kubernetes clusters in Rancher

martyav:
Starting a draft to help with checking that the links make sense and pages aren't missed


martyav:
In general, the pages in docs/how-to-guides/new-user-guides/kubernetes-clusters-in-rancher-setup/set-up-cloud-providers look like they need a refresh. There are style inconsistencies, formatting problems with putting headings in tabs, and language could be tightened up/clarified.

PR #779, opened 2023-08-15
Update rotate-certificates.md

donnyyu:
Add ca certs rotation via API procedures.

### DIFF
--- a/docs/how-to-guides/new-user-guides/manage-clusters/rotate-certificates.md
+++ b/docs/how-to-guides/new-user-guides/manage-clusters/rotate-certificates.md
@@ -62,6 +62,13 @@ Rancher launched Kubernetes clusters have the ability to rotate the auto-generat
 
 **Results:** The selected certificates will be rotated and the related services will be restarted to start using the new certificate.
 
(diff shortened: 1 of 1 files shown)


martyav (changes requested):

martyav (approved):

martyav:
There seems to be disagreement over whether instructions for using this endpoint should be included in the docs, as the endpoint might be removed in the future. Although I already approved, we might have to close this.

donnyyu:
Hi @martyav the only way to rotate ca is via the endpoint API call, the UI method doesn't have the option. Engineering team has confirmed it is the correct way to rotate ca. [SURE-6175]
You can remove the doc if endpoint is definitely removed from rancher, but it hasn't happened. My customer just wants to 
have an official guide that explains how to rotate ca

martyav:
@donnyyu Sorry to go back and forth on this. In general, instructions for using that endpoint are discouraged. But if rotating the CA can't be performed otherwise, then we might have to make an exception. I'm waiting to hear on how we should move forward.

PR #784, opened 2023-08-15
#420 SEO Improvements: pages-for-subheaders

martyav:
This adds canonical links to the the index pages in `pages-for-subheaders`. This is very welcome as they're among our more popular pages. However, the PR is very large since so many files are in that directory.


martyav:
For review's sake, I think this PR should be split, removing the "missing pages" commit to a separate branch. The missing pages (mostly 2.0-24, some 2.5) are the ones that really need scrutiny from a review.

martyav:
Review should focus on https://github.com/rancher/rancher-docs/pull/784/commits/23cbb2f1649098c984d0f0490771593983cad579 to make sure older pages are pointing to the correct latest version (correct in the sense that the older page no longer interferes with a newer page's SEO)

versioned_docs/version-2.0-2.4/pages-for-subheaders/rke-add-on.md
```diff
@@ -2,5 +2,9 @@
 title: RKE Add-On Install
 ---
 
+<head>
+  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/pages-for-subheaders/about-rke1-templates"/>
+</head>
+
```
btat, line 8:
```suggestion
```

I don't think there's an equivalent concept to link to in the latest docs. The first thing on the pages linked to on line 9 and 10 is `Important: RKE add-on install is only supported up to Rancher v2.0.8`. For now, let's leave such pages to self-reference until we find a better solution. In this case, this page is about using RKE to install Rancher whereas the proposed link is about downstream RKE clusters so using that as a canonical link is misleading.

versioned_docs/version-2.5/pages-for-subheaders/integrations-in-rancher.md
```diff
@@ -2,6 +2,10 @@
 title: Integrations in Rancher
 ---
 
+<head>
+  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/getting-started/overview"/>
```
btat, line 6:
```suggestion
  <link rel="canonical" href="https://ranchermanager.docs.rancher.com/pages-for-subheaders/cloud-marketplace"/>
```

Since our latest version has a sidebar category labeled exactly as "Integrations in Rancher", we point as close to that as possible. It's not a perfect solution, but cloud-marketplace is the first page under the Integrations category rather than a completely different area of the site.

Also, the 2.0-2.4 version of this page is missing from the PR.

btat (changes requested):

martyav:
@btat just pushed 2 commits addressing both comments. The page count remains the same because I removed the link in rke-add-ons and added a link to the 2.0-2.4 page for integrations-for-rancher

btat (approved):

PR #792, opened 2023-08-17
#787 - Update reference to outdated UI label

### LINKED ISSUE #787: Rancher CLI 2.7.  referring to 2.6
Hi team,
Our 2.7 documentation refers to 2.6 when it comes to install Rancher CLI, https://ranchermanager.docs.rancher.com/v2.7/reference-guides/cli-with-rancher/rancher-cli.

Besides that, this is for ranchermanager.docs, so I am unsure if this is the right place.

martyav:
Fixes #787 

It seems that the UI text for this link was changed at some point but we never updated the docs to account for it. This isn't the kind of thing to get release-noted so I'm not sure if it also affects some earlier minor versions of 2.6.x


martyav:
According to https://github.com/rancher/rancher-docs/issues/787#issuecomment-1683530232 this is a 2.7 exclusive

PR #794, opened 2023-08-22
Update rke-hardening-guide based on CIS-1.24 and CIS-1.7

andypitcher:
This PR updates the [RKE hardening guide](https://ranchermanager.docs.rancher.com/pages-for-subheaders/rke1-hardening-guide) with the new requirements added in `cis-1.24` and `cis-1.7`.
Related issue:
- https://github.com/rancher/rancher/issues/42154

### DIFF
--- a/docs/pages-for-subheaders/rke1-hardening-guide.md
+++ b/docs/pages-for-subheaders/rke1-hardening-guide.md
@@ -17,8 +17,8 @@ This hardening guide is intended to be used for RKE clusters and is associated w
 | Rancher v2.7    | Benchmark v1.7        | Kubernetes v1.25 up to v1.26 |
 
 :::note
-- Since Benchmark v1.24, check id `4.1.7 Ensure that the certificate authorities file permissions are set to 600 or more restrictive (Automated)` might fail, as /etc/kubernetes/ssl/kube-ca.pem is provisioned in 644 by default.
-- Since Benchmark v1.7 (latest), `--protect-kernel-defaults` (check id 4.2.6) parameter is not required anymore, and was replaced.
+- In Benchmark v1.24 and later, check id `4.1.7 Ensure that the certificate authorities file permissions are set to 600 or more restrictive (Automated)` might fail, as `/etc/kubernetes/ssl/kube-ca.pem` is set to 644 by default.
+- In Benchmark v1.7, the `--protect-kernel-defaults` (`4.2.6`) parameter isn't required anymore, and was removed by CIS.
 :::
 
 For more details on how to evaluate a hardened RKE cluster against the official CIS benchmark, refer to the RKE self-assessment guides for specific Kubernetes and CIS benchmark versions.
--- a/docs/pages-for-subheaders/rke1-hardening-guide.md
+++ b/docs/pages-for-subheaders/rke1-hardening-guide.md
@@ -295,6 +295,7 @@ services:
   kubelet:
     extra_args:
       feature-gates: RotateKubeletServerCertificate=true
+      protect-kernel-defaults: true
     generate_serving_certificate: true
 addons: |
   # Upstream Kubernetes restricted PSP policy
@@ -441,7 +442,6 @@ rancher_kubernetes_engine_config:
     kubelet:
       extra_args:
         feature-gates: RotateKubeletServerCertificate=true
-        protect-kernel-defaults: true
         tls-cipher-suites: TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256
       generate_serving_certificate: true
     scheduler:
--- a/docs/pages-for-subheaders/rke1-hardening-guide.md
+++ b/docs/pages-for-subheaders/rke1-hardening-guide.md
@@ -12,7 +12,9 @@ This hardening guide is intended to be used for RKE clusters and is associated w
 
 | Rancher Version | CIS Benchmark Version | Kubernetes Version           |
 |-----------------|-----------------------|------------------------------|
-| Rancher v2.7    | Benchmark v1.7       | Kubernetes v1.24 up to v1.25 |
+| Rancher v2.7    | Benchmark v1.23       | Kubernetes v1.23             |
+| Rancher v2.7    | Benchmark v1.24       | Kubernetes v1.24             |
+| Rancher v2.7    | Benchmark v1.7        | Kubernetes v1.25 up to v1.26 |
 
 :::note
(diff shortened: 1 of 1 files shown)


docs/pages-for-subheaders/rke1-hardening-guide.md
```diff
@@ -12,10 +12,11 @@ This hardening guide is intended to be used for RKE clusters and is associated w
 
 | Rancher Version | CIS Benchmark Version | Kubernetes Version           |
 |-----------------|-----------------------|------------------------------|
-| Rancher v2.7    | Benchmark v1.23       | Kubernetes v1.23 up to v1.25 |
+| Rancher v2.7    | Benchmark v1.7       | Kubernetes v1.24 up to v1.25 |
 
 :::note
-At the time of writing, the upstream CIS Kubernetes v1.25 benchmark is not yet available in Rancher. At this time Rancher is using the CIS v1.23 benchmark when scanning Kubernetes v1.25 clusters. Due to that, the CIS checks 5.2.3, 5.2.4, 5.2.5 and 5.2.6 might fail.
+- Since Benchmark v1.24, check id `4.1.7 Ensure that the certificate authorities file permissions are set to 600 or more restrictive (Automated)` might fail, as /etc/kubernetes/ssl/kube-ca.pem is provisioned in 644 by default.
+- Since Benchmark v1.7 (latest), `--protect-kernel-defaults` (check id 4.2.6) parameter is not required anymore, and was replaced.
```
martyav, line 19:
```suggestion
- In Benchmark v1.24 and later, check id `4.1.7 Ensure that the certificate authorities file permissions are set to 600 or more restrictive (Automated)` might fail, as /etc/kubernetes/ssl/kube-ca.pem is provisioned in 644 by default.
- In Benchmark v1.7, `--protect-kernel-defaults` (check id 4.2.6) parameter is not required anymore, and was replaced.
```

The danger in noting "latest" is that we'll have to remember to update when it's no longer latest.

martyav (changes requested):

docs/pages-for-subheaders/rke1-hardening-guide.md
andypitcher: Suggestion updated.

docs/pages-for-subheaders/rke1-hardening-guide.md
```diff
@@ -12,10 +12,11 @@ This hardening guide is intended to be used for RKE clusters and is associated w
 
 | Rancher Version | CIS Benchmark Version | Kubernetes Version           |
 |-----------------|-----------------------|------------------------------|
-| Rancher v2.7    | Benchmark v1.23       | Kubernetes v1.23 up to v1.25 |
+| Rancher v2.7    | Benchmark v1.7       | Kubernetes v1.24 up to v1.25 |
```
andypitcher, line 15: As per @Priyashetty17 suggestion.
```suggestion
| Rancher v2.7    | Benchmark v1.7       | Kubernetes v1.25 |
```

Priyashetty17: @andypitcher I did not realize this was the hardening guide. I got confused with the self assessment guide for v1.7. Since the upstream only supports cis v1.23 as of now, should we leave the CIS Benchmark Version in the Hardening Guide at 1.23 and K8s version "v1.23 up to v1.26"?

andypitcher: This is a good point, we could have one hardening-guide page for each distro, with one tab page for each CIS version (v1.23, v1.24 and v1.7) then, since cis-1.23 is the only option available upstream (it means we need to develop/provide cis-1.24 and cis-1.7 upstream).
@macedogm do you have some recommendation ?

JonCrowther:
You'll also need to make the changes in the versioned docs:
https://github.com/rancher/rancher-docs/blob/main/versioned_docs/version-2.7/pages-for-subheaders/rke1-hardening-guide.md

docs/pages-for-subheaders/rke1-hardening-guide.md
macedogm: @andypitcher I partially agree with the proposal:

> we could have one hardening-guide page for each distro

We already have this.

> with one tab page for each CIS version (v1.23, v1.24 and v1.7)

True. For the moment each tab is named as the K8s version, we would only have to change it to be named according to the CIS version too, perhaps from:

`v1.25 and Newer | v1.24 and Older`

To:

`CIS 1.7 (K8s v1.25) | CIS 1.24 (K8s v1.24) | CIS 1.23 / (K8s v1.22-v1.23)`

Is that what you mean?

image

> then, since cis-1.23 is the only option available upstream (it means we need to develop/provide cis-1.24 and cis-1.7 upstream).

I didn't get this part. Can you please elaborate further?

macedogm, line 15: Any reason why we are removing the versions < 1.25 from the hardening guide? I didn't get this change.
Rancher v2.7 still supports K8s v1.23 to v1.24, so we need to maintain them there.
Ideally the change should be something like this, unless I'm missing some context:

```suggestion
| Rancher v2.7    | Benchmark v1.23       | Kubernetes v1.23             |
| Rancher v2.7    | Benchmark v1.24       | Kubernetes v1.24             |
| Rancher v2.7    | Benchmark v1.7        | Kubernetes v1.25 up to v1.26 |
```

docs/pages-for-subheaders/rke1-hardening-guide.md
macedogm: @andypitcher I believe we only need to address this missing feedback from Marty, right?

> The danger in noting "latest" is that we'll have to remember to update when it's no longer latest.

docs/pages-for-subheaders/rke1-hardening-guide.md
macedogm: Thanks for the explanation. While the RKE2 flag isn't improved, should we just mention that users must use the `cis-1.23` for all CIS versions then?

docs/pages-for-subheaders/rke1-hardening-guide.md
```diff
@@ -243,7 +246,6 @@ services:
   kubelet:
     extra_args:
       feature-gates: RotateKubeletServerCertificate=true
-      protect-kernel-defaults: "true"
```
macedogm, line 246: This needs to be inside a specific tab depending on the CIS benchmark version, because it's still needed in CIS 1.23 and CIS 1.24.

docs/pages-for-subheaders/rke1-hardening-guide.md
```diff
@@ -293,7 +295,6 @@ services:
   kubelet:
     extra_args:
       feature-gates: RotateKubeletServerCertificate=true
-      protect-kernel-defaults: true
```
macedogm, line 296: Same as before:

> This needs to be inside a specific tab depending on the CIS benchmark version, because it's still needed in CIS 1.23 and CIS 1.24.

macedogm (changes requested):

docs/pages-for-subheaders/rke1-hardening-guide.md
macedogm: This flag was removed, but it still need to be kept because it's inside the v1.24 tab, no?

macedogm (changes requested):
@andypitcher two remaining items before approving this PR:

- `protect-kernel-defaults: true` is still present for v1.25 here -> https://github.com/rancher/rancher-docs/blob/4ca8cfada38f38908ba6e186fb874b5d8fbb275d/docs/pages-for-subheaders/rke1-hardening-guide.md?plain=1#L444. It needs to be removed.
- Addressing Marty's comment https://github.com/rancher/rancher-docs/pull/794/files#r1302017920.

macedogm (approved):

Issue #463, opened 2023-03-02
Update redirects due to hardening guide changes.


btat:
The updates https://github.com/rancher/rancher-docs/pull/453 renames/removes some existing hardening guide pages. Redirects should be added to direct user to the appropriate replacement page.

btat:
It seems we won't be able to add redirects for the hardening guide pages that have been updated. After some testing, it looks like the redirects plugin has a limitation where the last segment of a URL path in the `from` field of the redirect block doesn't allow periods in it. When there are periods in this part of the path, it will try to access the URL path as is and not redirect. Since the page no longer exists users will see `Page not found`. E.g. `/aa.a/bbb` will redirect but `/aaa/bb.b` will fail. 

These are the pages we'd be adding redirects for (the `from` field of the redirect block) and they all have periods in the last segment of the URL path, which uses the filename by default.

```
pages-for-subheaders/rancher-v2.6-hardening-guides
reference-guides/rancher-security/rancher-v2.6-hardening-guides/rke1-hardening-guide-with-cis-v1.6-benchmark
reference-guides/rancher-security/rancher-v2.6-hardening-guides/rke1-self-assessment-guide-with-cis-v1.6-benchmark
reference-guides/rancher-security/rancher-v2.6-hardening-guides/rke2-hardening-guide-with-cis-v1.6-benchmark
reference-guides/rancher-security/rancher-v2.6-hardening-guides/rke2-self-assessment-guide-with-cis-v1.6-benchmark
pages-for-subheaders/rancher-v2.7-hardening-guides
reference-guides/rancher-security/rancher-v2.7-hardening-guides/k3s-hardening-guide-with-cis-benchmark
reference-guides/rancher-security/rancher-v2.7-hardening-guides/k3s-self-assessment-guide-with-cis-v1.20-benchmark
reference-guides/rancher-security/rancher-v2.7-hardening-guides/k3s-self-assessment-guide-with-cis-v1.23-benchmark
```

PR #798, opened 2023-08-22
Update hardening guide links with broken redirects

martyav:
While testing a build, I noticed that the self-assessment guide redirects weren't resolving from pages-for-subheaders/rancher-hardening-guides.md. I updated the links on that page to point to the current names for those pages. When I checked the links in docusaurus.config, it seemed like the redirects were formatted correctly and pointed to the right paths. I know we had a similar issue some months ago with other redirects. Updating the links on pages-for-subheaders/rancher-hardening-guides.md is a quick fix -- we still need to look into why the redirects weren't resolving.


martyav:
Well...according to https://github.com/rancher/rancher-docs/issues/463#issuecomment-1494716941, the hardening guide pages have a known issue where periods within file names are incompatible with the redirects format. So, updating the links on the pages instead of  redirecting is the way to go.

PR #819, opened 2023-09-01
remove proxy_pass directive from server block (http) in nginx example config

AndrStnz:
The directive proxy_pass is not allowed in a server block with http.
Removing the proxy_pass directive at this location makes the sample useable.

See http module docs: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass


martyav:
@divya-mohan0209 can you have a look at this to confirm?

btat (approved):
Relevant bit from reference link indicating the valid context(s)

```
Context: location, if in location, limit_except
```

PR #823, opened 2023-09-06
clarified location of project  ID field

Tejeev:


martyav:
Looks good, just need to sync the file in /docs with the corresponding file in versioned_docs/version-2.7

PR #830, opened 2023-09-07
#829 Convert links to https://rancher.com/docs/rancher/v2.6 to use correct Markdown syntax

### LINKED ISSUE #829: Convert links to https://rancher.com/docs/rancher/v2.6 to use correct Markdown syntax
Within the docset, there are some lingering links to other docs pages, which don't use Markdown syntax. If an editor performs a search for all `https://` links expecting to return only links to external websites, they could be an annoyance, but they are otherwise harmless.

However, there are some pages in /docs and /v2.7 that point to `https://rancher.com/docs/rancher/v2.6`, so if a user clicks on them, they're taken to a docs page for the incorrect version of Rancher.

martyav:
Fixes #829 

> Within the docset, there are some lingering links to other docs pages, which don't use Markdown syntax. If an editor performs a search for all https:// links expecting to return only links to external websites, they could be an annoyance, but they are otherwise harmless.

> However, there are some pages in /docs and /v2.7 that point to https://rancher.com/docs/rancher/v2.6, so if a user clicks on them, they're taken to a docs page for the incorrect version of Rancher.


LucasSaintarbor (approved):
LGTM 👍

Issue #795, opened 2023-08-22
Refresh info for contributors


martyav:
We have a lot of information about contributing in our internal docs. It would be helpful for both internal and external contributors if some of the information was placed in this repo, as it would be easier to find and link to.

Some areas we could address:

- [x] The README needs more details about how to navigate the repo, edit versioned docs, and move pages. If it gets too long maybe we could have abbreviated instructions on the README covering basics, and another, more detailed file describing less-common tasks.
- [x] We currently have a page on Contributing here -- https://ranchermanager.docs.rancher.com/contribute-to-rancher. The information about docs is low on the page and points to the older 2.x repo. -- Addressed in #805 
- [x] We should have another look at our issue templates and update them as needed.

PR #833, opened 2023-09-08
#795 Update issue templates, add PR template

martyav:
Addresses this part of #795:

> We should have another look at our issue templates and update them as needed.

This PR adapts the format seen in the [release notes](https://raw.githubusercontent.com/rancherlabs/release-notes/main/.github/PULL_REQUEST_TEMPLATE.md?token=GHSAT0AAAAAACCYXKBERVJVCILWHU54YHIKZH3JVUQ) repo. Instructions are commented out so that the user can leave them in place when they start writing.

Also added a PR template. It encourages users to include an issue number, and reminds them about the release branch and versioning.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,13 +2,15 @@
 Check the Rancher docs issues to see if there is an existing issue for this pull request. If there is, enter the issue number below.
 -->
 
-Fixes #_
+Fixes #[issue_number]
 
 ## Reminder
 
+<!--
 Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.
 
 If the pull request is dependent on an upcoming release, make sure to target the release branch instead of `main`.
+-->
 
 ## Description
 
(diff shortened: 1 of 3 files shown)


sunilarjun (commented):
Just added some style updates that could be included but overall looks good, thanks!

.github/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
+<!--
+Check the Rancher docs issues to see if there is an existing issue for this pull request. If there is, enter the issue number below.
+-->
+
+Fixes #[issue_number]
+
+## Reminder
+
+<!--
+Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.
```
btat, line 10:
```suggestion
<!-- Reminder
Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.
```

I think we should make the `Reminder` label part of the comment. In the other instances where a header is used we're expecting the PR creator to fill in some information in the template, but that's not case here.

btat (changes requested):

btat (approved):

Issue #878, opened 2023-09-26
Undocumented "out-of-the box default global permissions"


lindhe:
I have two issues with the documentation regarding global permissions:

1. It says here that "Out-of-the-box, there are four default global permissions":

    https://github.com/rancher/rancher-docs/blob/c7c9caa3e64033ea05c09a7288f7f31e3fedb83f/versioned_docs/version-2.7/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/manage-role-based-access-control-rbac/global-permissions.md?plain=1#L11
    
    But there are clearly many more than that:
    
    <img width="775" alt="Global roles" src="https://github.com/rancher/rancher-docs/assets/7773090/e2e2bc86-8dce-4a1a-a459-dcc3b46facd4">

1. The documentation's nomenclature does not match the GUI. In the Rancher GUI, it's called "Global Roles", not "Global Roles", as far as I can tell. I think it would help the end users if you used a consistent terminology here.

lindhe:
I've now found that there is actually something called "global permissions" in the GUI, and it does match the list in the documentation. I found this when clicking "Edit config" for a user:

![Global Permissions](https://github.com/rancher/rancher-docs/assets/7773090/2db1496c-2ca9-4e4a-856d-2116a476365d)

How are they different than the global roles? This is very confusing to me.

lindhe:
Here, at least, the docs very clearly uses the term "global permissions" for one of the global roles:

> the **New User Default** global permissions

https://github.com/rancher/rancher-docs/blob/c7c9caa3e64033ea05c09a7288f7f31e3fedb83f/versioned_docs/version-2.7/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/manage-role-based-access-control-rbac/global-permissions.md?plain=1#L112

lindhe:
It says [here](https://github.com/rancher/rancher-docs/blob/c7c9caa3e64033ea05c09a7288f7f31e3fedb83f/versioned_docs/version-2.7/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/manage-role-based-access-control-rbac/custom-roles.md?plain=1#L11) that "Note that _roles_ are different from _permissions_, which determine what clusters and projects you can access."

This, to me, is very confusing considering what I've previously pointed out.

PR #881, opened 2023-09-26
Merge "Global Permission Assignment" subsections

lindhe:
## Reminders

## Description

This change consolidates the "Global Permission Assignment" instructions for local users and users who log in to Rancher using external authentication.

It said that "Global permissions for local users are assigned differently than users who log in to Rancher using external authentication", yet the instructions were identical (apart from some incorrect information, as far as I can tell).

## Comments

If you like these changes, I'll make the same change for the versioned docs too.

Relates to #878


lindhe:
I see that it's repeated yet again later on the same page: https://github.com/rancher/rancher-docs/blob/091ecbdde64219fcfa817b5a5f1cff821b042770/docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/manage-role-based-access-control-rbac/global-permissions.md#configuring-default-global-permissions

Should we remove this too? It seems like unnecessary work to put so many repeating statements in the same document...

PR #884, opened 2023-09-27
Merge main to 2023-Q3-v2.7x

btat:
Sync the release branch with latest updates.

PR #883, opened 2023-09-27
Merge 2023-Q3-v2.7x into main

btat:
To be merged at release time.

PR #762, opened 2023-08-08
#736 Update info related to EKS instance resource tags

martyav:
Fixes #736 

Also did a copyedit on that section as it was riddled with passive voice and the heading title wasn't capitalized according to current style guide standards.

It's tempting to do a style refresh on the rest of the page to match.

Including Richard Case as he was cc'd on the original issue.

PR #888, opened 2023-09-27
Merge 2023-Q3-v2.7x into main

btat:
To be merged at release time.


btat:
The incorrect base branch in #883 and #884 and subsequent revert in #886 seems like it's led to [this situation](https://stackoverflow.com/questions/49386213/reverted-a-pull-request-from-github-but-both-branch-shows-no-difference) resulting in the changes from #762 not being recognized. We'll need to reapply updates from #762 to main after the fact.

PR #966, opened 2023-10-31
Add 'More from SUSE' to header

btat:
## Description

Standardizes the header (placement of elements, More from SUSE, etc) to match other project docs sites. Also fixes the mixed usage of single and double quotes.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -14,14 +14,14 @@ module.exports = {
   projectName: 'rancher-docs', // Usually your repo name.
   trailingSlash: false,
   i18n: {
-    defaultLocale: "en",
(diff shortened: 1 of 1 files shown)


martyav:
Are we also adding the project icons like in Nuno's PR? https://github.com/rancher/elemental-docs/pull/235

btat:
> Are we also adding the project icons like in Nuno's PR? [rancher/elemental-docs#235](https://github.com/rancher/elemental-docs/pull/235)

The last time I spoke about this topic with Guna the plan was to use the plain version. I will check with Guna if we're moving to the style Nuno used as well as discussing the branding guidelines concern I brought up during the call last week.

PR #975, opened 2023-11-02
Colored icons for header

martyav:
## Comments

This largely takes code from https://github.com/rancher/elemental-docs/pull/235 (Thanks @nunix!)

I added the Elemental icon and tried to get a sharper version of the Rancher icon, as it looked blurry and stretched horizontally. The new icon is easier to see but cuts off half way, so it might need some CSS help. 

The PR also includes some unused icons for other projects, in case we want to add those to the menu (Rancher Desktop, Kubewarden, Opni). Epinio and Elemental include those project links -- but Epinio only has Kubewarden and Opni, while Elemental only has Rancher Desktop -- so all our sites have different links in the dropdown. Should we be consistent across all websites or pick the top 4 other projects associated with each project website?

image


btat:
> The PR also includes some unused icons for other projects, in case we want to add those to the menu (Rancher Desktop, Kubewarden, Opni).

IMO, we should only include the assets being used otherwise we just increase the repo size.

> Epinio and Elemental include those project links -- but Epinio only has Kubewarden and Opni, while Elemental only has Rancher Desktop -- so all our sites have different links in the dropdown. Should we be consistent across all websites or pick the top 4 other projects associated with each project website?
> 

This was previously discussed with @gunamata, but the list will always vary slightly because of the case where the given project site is in the list itself. The current implemented lists across project sites are based on the doc Guna provided.

src/css/custom.css
```diff
@@ -210,6 +211,75 @@ a.btn.navbar__github::before {
...
+  width: 20px;
+  margin-right: 4px;
+  background-color: var(--ifm-navbar-link-color);
+}
+
+.navbar__rancher:before {
+  mask: url(/static/img/header/icon-rancher.png) no-repeat 100% 100%;
+  mask-size: cover;
+  width: 35px;
+  padding-bottom: 7px;
```
btat, line 234:
```suggestion
  mask-size: contain;
  width: 35px;
```

Using `contain`:

header-contain

btat (changes requested):

btat (approved):

PR #926, opened 2023-10-13
Update the namespace exception list at multiple places 

jiaqiluo:
Rancher issues 
- https://github.com/rancher/rancher/issues/43078
- https://github.com/rancher/rancher/issues/42719

## Description

This PR is to add two new namespaces, `cattle-provisioning-capi-system` and `cattle-fleet-local-system`, to the exception namespace list which is mentioned at multiple places across the documentation. 

This PR also makes some general improvements:
- fix the wrong title
- add basic instructions for requirements for running rancher in hardened cluster 

## Comments

Those two new namespaces are added in rancher v2.7.7, therefore changes in this PR are applied to both the latest and versioned-v2.7 of the docs.

PR #1009, opened 2023-11-29
fix: update v2.7 cn sidebar

yzeng25:
## Description

The goal of this pull request is to fix the English title "Deploy Workloads" in V2.7's CN doc sidebar.

As checked, the variable is defined as `"label": "Deploy Workloads",` in the English V2.7 sidebar file `versioned_sidebars/version-2.7-sidebars.json` file.

I make changes to two things in `i18n/zh/docusaurus-plugin-content-docs/version-2.7.json`:
- change `sidebar.tutorialSidebar.category.Deploy Rancher Workloads` to `sidebar.tutorialSidebar.category.Deploy Workloads`, so that it can match with English V2.7 sidebar and display out the correct Chinese.
- update its description.

image

## Comments

I also checked the other sidebar.js files, V2.7 is the only one with `"label": "Deploy Workloads",`, while the other ones are `"label": "Deploy Rancher Workloads",`, including V2.8 and Preview. Not sure if changing the V2.7 English sidebar file back with `"label": "Deploy Rancher Workloads",` is ok.

### DIFF
--- a/i18n/zh/docusaurus-plugin-content-docs/current.json
+++ b/i18n/zh/docusaurus-plugin-content-docs/current.json
@@ -16,7 +16,7 @@
     "description": "The label for category Deploy Rancher in sidebar tutorialSidebar"
   },
   "sidebar.tutorialSidebar.category.Deploy Workloads": {
-    "message": "部署 Rancher 工作负载",
+    "message": "部署工作负载",
     "description": "The label for category Deploy Workloads in sidebar tutorialSidebar"
   },
   "sidebar.tutorialSidebar.category.Installation and Upgrade": {
--- a/i18n/zh/docusaurus-plugin-content-docs/current.json
+++ b/i18n/zh/docusaurus-plugin-content-docs/current.json
@@ -15,9 +15,9 @@
     "message": "部署 Rancher",
     "description": "The label for category Deploy Rancher in sidebar tutorialSidebar"
   },
-  "sidebar.tutorialSidebar.category.Deploy Rancher Workloads": {
+  "sidebar.tutorialSidebar.category.Deploy Workloads": {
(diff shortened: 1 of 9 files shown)


martyav:
I checked the other sidebars, and there are some odd inconsistencies. @btat 

- The label in the sidebar file for /docs uses single quotes for 'Deploy Workloads'
- The sidebar file in v2.7 uses double quotes for "Deploy Workloads"
- The sidebar file in v2.8 uses double quotes and calls that section *Deploy Rancher Workloads*

So, it seems like the label was updated some time after v2.8 was split off. Do we want to use the shorter version, or the longer one, across all versions?

btat:
> I checked the other sidebars, and there are some odd inconsistencies. @btat
> 
>     * The label in the sidebar file for /docs uses single quotes for 'Deploy Workloads'
> 
>     * The sidebar file in v2.7 uses double quotes for "Deploy Workloads". That's odd because it should be a copy of the previous file
> 
>     * The sidebar file in v2.8 uses double quotes and calls that section _Deploy Rancher Workloads_
> 
> 
> So, it seems like the label was updated some time after v2.8 was split off from /docs. Do we want to use the shorter version, or the longer one, across all versions?

- Quoting: for the top-level sidebars.js we follow Docusaurus' convention of using single quotes. As for the generated versioned sidebar files, the files are in JSON format, which requires double quotes to be valid. See https://www.json.org.
- Label: the label was updated as part of https://github.com/rancher/rancher-docs/pull/926 and should've been synced to other versions. My interpretation of the change is that the previous label could be misinterpret as "workloads for Rancher".

btat (approved):

martyav (approved):

btat:
Thanks for the PR @yzeng25

PR #1021, opened 2023-12-06
Remove unreleased banner

btat:


martyav:
Removing the unreleased banner inadvertently applied the deprecation banner. The deprecation banner needs to be explicitly opted out of, either by overriding it with another banner or specifying `none`: https://github.com/rancher/rancher-docs/pull/1023

Issue #991, opened 2023-11-20
Refresh and restructure cloud provider guide section


martyav:
## Summary

As noted in https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-clusters-in-rancher-setup/set-up-cloud-providers/migrate-from-in-tree-to-out-of-tree, "Kubernetes is moving away from maintaining cloud providers in-tree."

However, we currently have only one dedicated page for out-of-tree migration, and it's specifically for VSphere. Similarly, there's only one page for out-of-tree setup that is specific only to VSphere. AWS content about out-of-tree setup and migration is hosted on a single AWS page. 

I think, in the interests of page length, searchability, and maintainability, we should have separate pages for each set of instructions for each provider.

@jakefhyde [proposes](https://github.com/rancher/rancher-docs/pull/844#issuecomment-1817233154), "Personally, I think we should restructure the docs to have nested folders for Amazon, Azure, GCE, and VSphere, and within those have separate pages for in-tree, out-of-tree, and the migration case."

The comment also suggests adding more detail to the instructions. Much of our instructions rely on the UI and don't provide commands for users working from the CLI.

martyav:
From Slack:

> AWS cloud provider docs are definitely looking better than the state I left them in so I thank you for that, I do have a suggestion though. On [https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-clusters-in[…]-of-tree-cloud-provider/migrate-to-out-of-tree-amazon](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-clusters-in-rancher-setup/migrate-to-an-out-of-tree-cloud-provider/migrate-to-out-of-tree-amazon), the following line:

>> If it's acceptable to have some downtime, you can [switch to an external cloud provider](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/kubernetes-clusters-in-rancher-setup/set-up-cloud-providers/amazon#using-the-out-of-tree-aws-cloud-provider), which removes in-tree components and then deploy charts to install the AWS cloud controller manager.

> reads like it's optional. I think I wrote it initially but now that I'm looking at it with fresh eyes, it doesn't really make sense. The link is for setting up the external cloud-provider on a fresh cluster, so I feel like the sentence should read something like:

[...]

PR #844, opened 2023-09-14
Add aws out of tree cloud provider install/upgrade docs

jakefhyde:
Fixes #763 

I copied a lot of what Kinara wrote in https://github.com/rancher/rancher-docs/pull/826 wholesale.

This PR is ready for review and supersedes https://github.com/rancher/rancher-docs/pull/826.

For https://github.com/rancher/rancher/issues/42749

PR #1025, opened 2023-12-08
#991 refresh and restructure cloud provider guide section PART ONE: Restructure

martyav:
Part one of #991: Restructure

## Reminders

## Description

This moves files in the /set-up-cloud-providers directory, creates two new subfolders for Amazon and vSphere respectively, and  addresses some broken links caused by the path changes.

## Comments

We should still refresh the pages in this section with CLI instructions and fill in any knowledge gaps.

When you review, you should build the site locally and preview it to make sure the pages are in the right place and links are working. I checked on my local machine before submitting the PR and didn't find anything odd.


btat (changes requested):
Functional changes look good, but with the proposed design

- The naming scheme for files on the same topic (just for different providers) should be updated to match so it's obvious that they're about the same topic.

- Files for topics on the same topic for different providers ends up on different levels, which looks a bit odd when looking at the sidebar. I'm not sure what we should do about the vSphere ones, but at least for the Amazon instance `configure-amazon` should be the category instead of a child page.


Also, what's the plan for part two?

martyav:
For part 2, we should review the other pages in this section and see if we need to include more detailed instructions to match the current Amazon page. We might also need to revise the instructions for clarity and consistency. When we determine the changes needed, we should work with an SME to update the pages (which commands to add, how to validate that something succeeded).

from https://github.com/rancher/rancher-docs/pull/844#issuecomment-1817233154:

> I agree with the feedback you've given on this PR in regards to including manual commands as well as validation steps, and looking at the documentation on other cloud providers, I do not feel like they meet the same level of depth.

sunilarjun (commented):
Smoke tested the pages/links and everything looks good! I do agree that the file naming scheme should be similar for grouped topics.

martyav:
Recently pushed changes to move the pages out of the deeply nested directories and fix links

btat (changes requested):
The [latest build](https://github.com/rancher/rancher-docs/actions/runs/7225219412/job/19688166053?pr=1025) still shows a handful of broken links related to the files from this PR.

btat (approved):

PR #1024, opened 2023-12-07
Link fixes

btat:
## Description

This PR fixes links broken due to various reasons (renamed header, missing file extension, file link to the document itself, etc.) and adds missing group IDs to some tabs.

PR #1027, opened 2023-12-08
Remove file path from link to section within same file

btat:


martyav:
Are you going to add any other pages, besides `use-new-nodes`, to the pull request? As is, it looks good.

btat:
> Are you going to add any other pages, besides `use-new-nodes`, to the pull request? As is, it looks good.

There was another one, but turns out I already fixed that in https://github.com/rancher/rancher-docs/pull/1024.

PR #1033, opened 2023-12-12
Add `cloudcredential` as question type

lindhe:
## Reminders

## Description

I tried [this example](https://github.com/rancher/cluster-template-examples/blob/dc5c201c5415c430cdf4bcb6af975f9e516e970b/charts/questions.yaml#L13) and found that `cloudcredential` is indeed a valid type. It seems to be a drop-down menu that selects among the user's available Cloud Credentials.

image


lindhe:
I want to hear if you approve of this change before I take the time and edit the versioned docs too.

btat (approved):
LGTM

lindhe:
That's the rest of them updated. I just realized that maybe not all versions support that type? What do you think? I can't test that.

lindhe:
Ah, excellent finding! I'll update the PR accordingly.

lindhe:
Should be good now, I hope! 👍

btat:
Looks good, thanks!

PR #1067, opened 2024-01-16
Indicate distros mentioned aren't an exhaustive list

btat:
## Description

Per feedback, the phrasing used to make call-outs for RKE and K3s were causing confusion about the ability to perform the procedure with other distros.

Also fixed the outdated URLs.


gunamata (approved):
LGTM, Thank you.

Issue #1069, opened 2024-01-17
v2.8.1 release maintenance checklist


btat:
This issue is to track tasks that need to be done every release regardless of whether the release has new feature content or not.

- [x] Update the [versions table](https://ranchermanager.docs.rancher.com/versions) (#1072)
- [x] Update the [Rancher:webhook version mapping table](https://ranchermanager.docs.rancher.com/reference-guides/rancher-webhook) (#1070)
- [x] Update the [CNI popularity table](https://ranchermanager.docs.rancher.com/faq/container-network-interface-providers#cni-community-popularity) (#1071)
- [x] Create a new [release](https://github.com/rancher/rancher-docs/releases) (https://github.com/rancher/rancher-docs/releases/tag/initial-snapshot-v2.8.1)
- [x] Update Algolia search index

PR #1072, opened 2024-01-17
Add 2.8.1 entry to versions table

btat:
Related to #1069 

## Description

Adds an entry for 2.8.1.

Also for versions that don't have an entry on the support matrix, the link is replaced with "N/A".

### DIFF
--- a/src/pages/versions.md
+++ b/src/pages/versions.md
@@ -10,10 +10,10 @@ Below are the documentation and release notes for the currently released version
 
 <table>
   <tr>
-    <td><b>v2.8.0</b></td>
+    <td><b>v2.8.1</b></td>
     <td><a href="https://ranchermanager.docs.rancher.com/v2.8">Documentation</a></td>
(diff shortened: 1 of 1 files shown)


src/pages/versions.md
```diff
@@ -41,20 +41,31 @@ Below are the documentation and release notes for the currently released version
 
 ### Past versions
 
-Below are the documentation and release notes for previous versions of Rancher 2.7.x:
```
sunilarjun, line 44: I think this line is still needed below to denote the Rancher 2.7.x release items, wasn't sure where to add in below but wanted to make known.

sunilarjun (commented):
Just left a note that the line `Below are the documentation and release notes for previous versions of Rancher 2.7.x:` is missing above the 2.7.x release items, otherwise looks good!

src/pages/versions.md
btat: Thanks for catching this Sunil. Fixed in 7cb977a7.

martyav (approved):

sunilarjun (approved):

PR #1126, opened 2024-02-15
Webhook table - indicate Prime/Community availability

btat:
## Description

Per discussion with Guna, we should add additional columns to indicate availability of a given version.

### DIFF
--- a/docs/reference-guides/rancher-webhook.md
+++ b/docs/reference-guides/rancher-webhook.md
@@ -17,11 +17,11 @@ Each Rancher version is designed to be compatible with a single version of the w
 
 <!-- releaseTask -->
 
-| Rancher Version | Webhook Version |
-|-----------------|:---------------:|
-| v2.8.0          |     v0.4.2      |
-| v2.8.1          |     v0.4.2      |
-| v2.8.2          |     v0.4.2      |
+| Rancher Version | Webhook Version | Prime   | Community |
+|-----------------|-----------------|---------|-----------|
+| v2.8.2          |     v0.4.2      | &check; | &check;   |
+| v2.8.1          |     v0.4.2      | &check; | &check;   |
+| v2.8.0          |     v0.4.2      |         | &check;   |
 
 ## Why Do We Need It?
 
--- a/docs/reference-guides/rancher-webhook.md
+++ b/docs/reference-guides/rancher-webhook.md
@@ -17,11 +17,11 @@ Each Rancher version is designed to be compatible with a single version of the w
 
 <!-- releaseTask -->
 
-| Rancher Version | Webhook Version | Prime   | Community |
-|-----------------|-----------------|---------|-----------|
-| v2.8.2          |     v0.4.2      | &check; | &check;   |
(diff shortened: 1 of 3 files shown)


docs/reference-guides/rancher-webhook.md
```diff
@@ -17,11 +17,11 @@ Each Rancher version is designed to be compatible with a single version of the w
...
-| Rancher Version | Webhook Version |
-|-----------------|:---------------:|
-| v2.8.0          |     v0.4.2      |
-| v2.8.1          |     v0.4.2      |
-| v2.8.2          |     v0.4.2      |
+| Rancher Version | Webhook Version | Prime   | Community |
+|-----------------|-----------------|---------|-----------|
+| v2.8.2          |     v0.4.2      | &check; | &check;   |
+| v2.8.1          |     v0.4.2      | &check; | &check;   |
+| v2.8.0          |     v0.4.2      |         | &check;   |
```
martyav, line 24: I think we should follow https://ranchermanager.docs.rancher.com/versions and fill the empty cells with text (N/A, Not supported, etc.) or a symbol (X). Otherwise, the empty cells might seem like a mistake, or like we may add availability in the future.

docs/reference-guides/rancher-webhook.md
```diff
@@ -17,11 +17,11 @@ Each Rancher version is designed to be compatible with a single version of the w
 
 <!-- releaseTask -->
 
-| Rancher Version | Webhook Version |
-|-----------------|:---------------:|
-| v2.8.0          |     v0.4.2      |
-| v2.8.1          |     v0.4.2      |
-| v2.8.2          |     v0.4.2      |
+| Rancher Version | Webhook Version | Prime   | Community |
```
martyav, line 20: The first two columns indicate compatibility between versions of different things (Rancher and Webhook). The next two columns indicate if a Rancher version is available through one thing (Prime) or another thing (Community). Suggest renaming column headers to something like "Available in Prime/Community" to make it clear that the last two columns are describing a different relationship than the first two.

docs/reference-guides/rancher-webhook.md
martyav: We might also want to include notes (in the first column or as footnotes) to explain why something isn't available (unstable release, Prime exclusive, etc)

btat: Added crosses (x) instead of empty cells in 53357bc7f06fbb67d99f5be7c3a51c6cd866e090.

> explain why something isn't available

I didn't add this since there's no direct relation with the webhook.

docs/reference-guides/rancher-webhook.md
btat: Updated in https://github.com/rancher/rancher-docs/commit/53357bc7f06fbb67d99f5be7c3a51c6cd866e090.

martyav (approved):

sunilarjun (approved):

LucasSaintarbor (approved):

PR #1127, opened 2024-02-15
Versions table - indicate Prime/Community availability

btat:
## Description

Per discussion with Guna, we should add additional columns to indicate availability of a given version.

### DIFF
--- a/src/pages/versions.md
+++ b/src/pages/versions.md
@@ -10,10 +10,20 @@ Below are the documentation and release notes for the currently released version
 
 <table>
   <tr>
+    <th>Version</th>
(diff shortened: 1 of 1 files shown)


martyav (commented):
The formatting looked fine in preview. It feels like we should open with a blurb explaining that Rancher Prime is only available for 2.8 and 2.7, that Community is the free tier, and that Community corresponds to the Rancher (edition? software line?) available for 2.6 and below. Otherwise, customers might have questions about what the columns mean and what they're getting.

martyav (approved):

LucasSaintarbor (approved):

PR #1186, opened 2024-03-20
Update to add additional linked libraries to be mounted as a volume to run `systemd-run`

chiukapoor:
### Issue:

* Fixes https://github.com/rancher/rancher/issues/44203

* To fix the [GlusterFS kubelet image](https://github.com/rancher/rancher/issues/40508) issue we mounted the additional linked libraries as volume from host to Hyperkube Kubelet container along with `systemd-run`, `-v /usr/lib/x86_64-linux-gnu/libcrypto.so.3:/usr/lib/x86_64-linux-gnu/libcrypto.so.3` and `-v /lib/systemd/l
ibsystemd-shared-249.so:/lib/systemd/libsystemd-shared-249.so`

* Command and manifest of [GlusterFS Volumes](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/manage-clusters/create-kubernetes-persistent-storage/manage-persistent-storage/about-glusterfs-volumes) document is updated with the following

```
docker run -v /usr/bin/systemd-run:/usr/bin/systemd-run -v /usr/lib/x86_64-linux-gnu/libcrypto.so.3:/usr/lib/x86_64-linux-gnu/libcrypto.so.3 -v /lib/systemd/l ibsystemd-shared-249.so:/lib/systemd/libsystemd-shared-249.so --entrypoint /usr/bin/systemd-run rancher/hyperkube:v1.28.7-rancher1 --version
```

```
services:
  kubelet:
    extra_binds:
      - "/usr/bin/systemd-run:/usr/bin/systemd-run"
      -  "/usr/lib/x86_64-linux-gnu/libcrypto.so.3:/usr/lib/x86_64-linux-gnu/libcrypto.so.3"
      -  "/lib/systemd/l ibsystemd-shared-249.so:/lib/systemd/libsystemd-shared-249.so"
```

### Related
* https://github.com/rancher/hyperkube-base/pull/20
* https://github.com/rancher/rancher/issues/40508
* https://github.com/rancher/rancher/issues/41601

### DIFF
--- a/docs/how-to-guides/new-user-guides/manage-clusters/create-kubernetes-persistent-storage/manage-persistent-storage/about-glusterfs-volumes.md
+++ b/docs/how-to-guides/new-user-guides/manage-clusters/create-kubernetes-persistent-storage/manage-persistent-storage/about-glusterfs-volumes.md
@@ -18,7 +18,7 @@ In clusters that store data on GlusterFS volumes, you may experience an issue wh
 - The `systemd-run` binary needs to be compatible with Debian OS on which the hyperkube image is based (this can be checked using the following command on each cluster node, replacing the image tag with the Kubernetes version you want to use)
 
 ```
-docker run -v /usr/bin/systemd-run:/usr/bin/systemd-run --entrypoint /usr/bin/systemd-run rancher/hyperkube:v1.16.2-rancher1 --version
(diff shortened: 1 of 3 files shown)


snasovich:
> Do these instructions apply to Rancher v2.7 and earlier, or just v2.8? Asking as we have separate files for Rancher v2.0-2.4, 2.5, etc. that would need to be updated as well.

I believe it should be updated for 2.7 as well (maybe even 2.6) - but given 2.6 is practically EoL I think it's fine to update only 2.8/2.7.

Though for 2.7 we want to make sure to not reference K8s 1.28 as it's not supported there. (FWIW it's not yet supported for 2.8 yet until 2.8.3 release so maybe we can reference earlier K8s version in all the cases?)
I understand it's just an example but it could be safer to just reference K8s 1.26 on these examples.

@chiukapoor , please update accordingly.

chiukapoor:
Updated the versioned docs for 2.7 and 2.8 with 1.26.14 k8s reference.

snasovich (approved):

PR #1191, opened 2024-03-22
Update trobleshooting tips regarding leader election

aruiz14:
Starting on [Rancher 2.8.3](https://github.com/rancher/rancher/pull/43652) and 2.9.X, only `Lease` objects will be used as a lock for leader election. Previously, a multi-lock `ConfigMap` + `Lease` was used, so these docs changes are backward compatible.


martyav (approved):
Looks good, just need to wait until v2.8.3 is out until we merge.

PR #1192, opened 2024-03-25
Fix broken canonical links for PAYG

martyav:
Fixes #[issue_number]

## Reminders

## Description

I noticed that the canonical links on these files didn't lead anywhere, so I updated them to point to the correct URLs.

## Comments

- PAYG is for v2.7.9, so we don't have corresponding files yet in `/docs` or `version-2.8`.

- The other pages in these two subdirectories don't have manual canonical links yet. We add manual canonical links to override the default Docusaurus behavior, which is to have each file's canonical link point to the URL for that file's webpage. Example: Docusaurus makes the canonical link for v2.5 pages point to `https://ranchermanager.docs.rancher.com/v2.5/`. 

- Since the other files in the PAYG subdirectories don't have any corresponding versioned files yet, Docusaurus currently generates, correctly, canonical links that point to the files' webpages in `https://ranchermanager.docs.rancher.com/v2.7/` 

- The alternative way to fix the broken links on these pages is by removing lines 5-7 and not setting manual canonical links until we have to add versioned files for `v2.8` and `/docs`. What do you think?

### DIFF
--- a/versioned_docs/version-2.7/integrations-in-rancher/cloud-marketplace/aws-marketplace-payg-integration/aws-marketplace-payg-integration.md
+++ b/versioned_docs/version-2.7/integrations-in-rancher/cloud-marketplace/aws-marketplace-payg-integration/aws-marketplace-payg-integration.md
@@ -2,10 +2,6 @@
 title: AWS Marketplace Pay-as-you-go (PAYG) Integration
 ---
 
-<head>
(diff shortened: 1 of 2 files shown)


sunilarjun (approved):

btat (commented):
> The alternative way to fix the broken links on these pages is by removing lines 5-7 and not setting manual canonical links until we have to add versioned files for v2.8 and /docs. What do you think?

I think either approach is fine as long as it's consistent across all files in the PAYG directories and not a mixture like now.

btat (approved):

PR #1111, opened 2024-02-08
#1108 Migrating Rancher to a new cluster lists old chart version

martyav:
Fixes #1108

## Reminders

- See the [README](../README.md) for more details on how to work with the Rancher docs.

- Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.

- If the pull request is dependent on an upcoming release, make sure to target the release branch instead of `main`.

## Description

From a contributor:

> 2.x.x is old and latest is 103.0.0+up4.0.0. Assuming latest should be used, the docs need to be updated to reflect this.

Initially edited to reflect this suggestion, then thought better of it, and linked users to the support matrix. I removed the specific mention of version numbers because that is going to continuously change. 

## Comments

I did some investigating to confirm. From our release notes:

> Rancher Helm Chart Versions
> Starting in 2.6.0, many of the Rancher Helm charts available in the Apps & Marketplace will start with a major version of 100. This was done to avoid simultaneous upstream changes and Rancher changes from causing conflicting version increments. This also brings us into compliance with semver, which is a requirement for newer versions of Helm. You can now see the upstream version of a chart in the build metadata, for example: 100.0.0+up2.1.0. See https://github.com/rancher/rancher/issues/32294.

The issue the release notes link to is from 2021 -- and specifically notes that rancher-backup wasn't included in the updated numbering system: https://github.com/rancher/rancher/issues/32294#issuecomment-906006333

However, at some point **after** 2021, this was changed, and this is confirmed both by the [support matrix](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-8-1/) and by running `helm search repo --versions rancher-charts/rancher-backup` to view available versions. 

Note that the [support matrix for v2.6](https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-6-13/) states that **the latest version of v2.6 is still using backup charts in the 2.x.x line**.

<s>Creating as a draft as I still need to add some lines about how the user can refer to the support matrix to confirm compatibility with their version of Rancher.</s>

PR #1193, opened 2024-03-25
#911 migration doc in need of clarification

### LINKED ISSUE #760: "Migrating Rancher to a New Cluster" lacks note about redirecting to new cluster
From Slack:

> I noticed that https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster no longer notes that you should redirect your DNS entry or load balancer to the location of the new cluster... Is that somehow not a thing any more?

>> There's the blurb in the caution box in the [prereqs section](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster#prerequisites), maybe that's what you're thinking of?

> The note seems to brush up against it but not address the actual technical requirement as I know them (Keep FQDN/URL the same and refer to docs on how to change them if needed). The Note seems to say to keep the node hostname the same but I don't remember that being important (perhaps that's  changed and i'm just remembering something like Rancher 2.1

### LINKED ISSUE #911: Migration doc is in need of clarification
## Summary
https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster

- It seems the following line means to refer to `<2.x.x>` to mean a variable that can equal any rancher 2 minor and dot version.  If this is the case, then it should follow the same practice as other variables by stating the variable name, what values it can take, then providing the example with the variable in place of the value
```
Select and set CHART_VERSION variable with a 2.x.x rancher-backup release version:

helm search repo --versions rancher-charts/rancher-backup
CHART_VERSION=<2.x.x>
```

- in that same theme, the following doesn't actually say that the `$REGISTRY` variable should be replaced with their path to registry
```
For an air-gapped environment, use the Helm value below to pull the backup-restore-operator image from your private registry when installing the rancher-backup Helm chart.

--set image.repository $REGISTRY/rancher/backup-restore-operator
```

- There is no mention of redirecting DNS after the migration so that it points to the new cluster.  This should be the last step.

martyav:
Fixes #911
Fixes #760 

## Reminders

## Description

From the issue:

> https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster
> 
> - It seems the following line means to refer to `<2.x.x>` to mean a variable that can equal any rancher 2 minor and dot version.  If this is the case, then it should follow the same practice as other variables by stating the variable name, what values it can take, then providing the example with the variable in place of the value
> ```
> Select and set CHART_VERSION variable with a 2.x.x rancher-backup release version:
> 
> helm search repo --versions rancher-charts/rancher-backup
> CHART_VERSION=<2.x.x>
> ```

The above point was already handled by #1111, *Migrating Rancher to a new cluster lists old chart version*, so it isn't part of this PR.

> - in that same theme, the following doesn't actually say that the `$REGISTRY` variable should be replaced with their path to registry
> ```
> For an air-gapped environment, use the Helm value below to pull the backup-restore-operator image from your private registry when installing the rancher-backup Helm chart.
> 
> --set image.repository $REGISTRY/rancher/backup-restore-operator
> ```

This point is included in the PR, as a slight update to wording.

> - There is no mention of redirecting DNS after the migration so that it points to the new cluster.  This should be the last step.

This point is also addressed here, by adding a section for the final step. I thought about adding links to Azure, AWS, GCP etc. docs on DNS, but decided not to, since then we would have to make judgements about who gets on the list, update dead links, etc.


martyav:
The comment on #760 notes:

> The note seems to brush up against it but not address the actual technical requirement as I know them (Keep FQDN/URL the same and refer to docs on how to change them if needed). The Note seems to say to keep the node hostname the same but I don't remember that being important (perhaps that's changed and i'm just remembering something like Rancher 2.1)

docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
```diff
@@ -15,7 +15,7 @@ These instructions assume that you have [created a backup](back-up-rancher.md) a
 
 :::caution
 
-It is required to use the same hostname that was set as the server URL in the first cluster. If not done, downstream clusters will show as unavailable in the cluster management page of the UI, and you won't be able to click inside the cluster or on the cluster's <b>Explore</b> button.
+You must use the same hostname that was set as the server URL in the original cluster. If you don't, downstream clusters will show as unavailable in the cluster management page of the UI, and you won't be able to click inside the cluster or on the cluster's <b>Explore</b> button.
```
btat, line 18:
```suggestion
You must use the same hostname that was set as the server URL in the original cluster. If you don't, downstream clusters will show as unavailable in the cluster management page of the UI, and you won't be able to click inside the cluster or on the cluster's **Explore** button.
```

versioned_docs/version-2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
```diff
@@ -52,7 +52,7 @@ Install the [`rancher-backup chart`](https://github.com/rancher/backup-restore-o
 
      The above assumes an environment with outbound connectivity to Docker Hub.
 
-     For an **air-gapped environment**, use the Helm value below to pull the `backup-restore-operator` image from your private registry when installing the rancher-backup Helm chart.
+     For an **air-gapped environment**, set a `$REGISTRY` variable to point to the path to your registry. Then, use the following Helm value to pull the `backup-restore-operator` image from your private registry when you install the rancher-backup Helm chart.
 
      ```bash
      --set image.repository $REGISTRY/rancher/backup-restore-operator
```
btat, line 58:
```suggestion
     For an **air-gapped environment**, use the following `image.repository` Helm value to pull the `backup-restore-operator` image from your private registry when you install the rancher-backup Helm chart. Replace `<registry>` with  your registry URL.

     ```bash
     --set image.repository <registry>/rancher/backup-restore-operator
```

I'm thinking `$REGISTRY` in the original example command was intended to be a placeholder and not an actual variable? There's only a single occurrence on this page so using a variable adds extra steps to do the same thing. If we want to keep it a variable though, then the example should be like what line 43 did and have

```
REGISTRY=<registry>
```

btat (changes requested):

versioned_docs/version-2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
martyav: Probably just simpler to use `<registry>`

martyav:
> Just a minor change but lines 11 and 28 can be deleted for markdown syntax alignment, otherwise outside of above comments changes look good. Thanks Marty!

@sunilarjun can you clarify which file these lines are on? Thanks!

sunilarjun:
> > Just a minor change but lines 11 and 28 can be deleted for markdown syntax alignment, otherwise outside of above comments changes look good. Thanks Marty!
> 
> @sunilarjun can you clarify which file these lines are on? Thanks!

Yeah my bad!

- `docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md`
- `versioned_docs/version-2.8/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md`
- `versioned_docs/version-2.7/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md`
  - Lines 11 and 28 can be deleted as they are extra new lines

sunilarjun (approved):

btat (approved):

PR #1217, opened 2024-04-05
Update general-faq.md

bgulla:
removing notice about mesos & swarm. it's been 6+ years since 2.0 was released, time to move on.

Fixes #[issue_number]

## Reminders


martyav:
I would suggest also removing from `versioned_docs/version-2.6/faq.md`, `versioned_docs/version-2.7/faq/faq.md` and `/version-2.8/faq/faq.md`. We can leave it in the earliest versioned docs files, since that's where it would most likely come up as a question (if it comes up as a question).

martyav (approved):

Issue #593, opened 2023-05-05
Add password requirements


btat:
Creating this issue as an extension of https://github.com/rancher/rancher-docs/pull/588.

The password requirements are provider agnostic so a better solution is to list out the password requirements in a central area and link to it rather than repeating them on each provider's page.

Possible pages affected:
- https://ranchermanager.docs.rancher.com/getting-started/installation-and-upgrade/resources/bootstrap-password

Related:
- https://github.com/rancher/rancher-docs/pull/592

martyav:
Some relevant PRs

---

## Parent issue

- https://github.com/rancher/dashboard/issues/6633
---

## Pull requests

- https://github.com/rancher/dashboard/pull/6801 - v2.6.9:
  - Added new setting inside "Global Setting" using key password-min-length.
- https://github.com/rancher/dashboard/pull/6897 - v2.7.2:
  - Prevent users to submit password configuration which is not matching the requirement: min 2 and max 256 chars. 
- https://github.com/rancher/dashboard/pull/7930 - v2.7.2: 
  - Added validation for integers exclusively for min-password changes
  - Changed setting key CATTLE_PASSWORD_MIN_LENGTH to PASSWORD_MIN_LENGTH, to match server returned value
  - Allowed editing settings from environment variables; this is due server configuration have been changed for this case
  - Updated tests - https://github.com/rancher/dashboard/pull/7930/commits/3d5a9810e7dfc2e115e38072f0e045ff3d60bf77
- https://github.com/rancher/dashboard/pull/8006 - v2.7.2
  - New case with decimal using 0 as value, e.g. 1.0.

PR #1208, opened 2024-04-01
#593 add password requirements

martyav:
Fixes #593

## Reminders

- See the [README](../README.md) for more details on how to work with the Rancher docs.

- Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.

- If the pull request is dependent on an upcoming release, make sure to target the release branch instead of `main`.

## Description

This adds the password requirements to a central location and refreshes the **Setting a Bootstrap Password** page to use tabs and remove redundant material.

Also: Users are not actually REQUIRED to reset the admin password after first login, though they likely should.

## Comments

Backticks don't seem to work inside tabs unless they're at the end of a line or on their own line?

A page in the dashboard repo makes some claims about passwords that I'm trying to verify: https://github.com/rancher/dashboard/blob/master/docusaurus/docs/code-base-works/auth-sessions-and-tokens.md

1. ~The minimum password length can be modified by the user (not sure at all if this is relevant to the bootstrap password)~ This doesn't apply to the bootstrap password. Passwords made later must be 12 chars or longer.
2. Password and username must differ

PR #1219, opened 2024-04-05
#1213 Instructions on altering minimum password length

### LINKED ISSUE #1213: Instructions on altering minimum password length
## Related Issues

https://github.com/rancher/rancher-docs/issues/593

https://github.com/rancher/dashboard/issues/6633

## Summary

By default, the minimum password length for accounts is 12 chars. 

Rancher v2.6.9 introduced the ability to update the minimum length from Global Settings in the UI.

Rancher v2.7.2 introduced further restrictions: password length must be a positive integer (or a string coerced into a positive integer value) between 2 and 256 chars long. Leading zeroes in octal values are allowed. Decimal values are not allowed (1 is valid, 01 is valid, 1.0 is not). 

We should describe how to alter the minimum password length and the associated restrictions.

martyav:
Fixes #1213
Related to #593 

## Reminders

## Description

This describes how to modify a field that was added to Global Settings in v2.6: https://github.com/rancher/dashboard/pull/6801

### DIFF
--- a/docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/manage-users-and-groups.md
+++ b/docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/manage-users-and-groups.md
@@ -72,7 +72,7 @@ Since SAML does not support user lookup, SAML-based authentication providers do 
 
 ## Minimum Password Length
 
-By default, user passwords must be at least 12 characters long. However, you can customize the password length requirment:
+By default, user passwords must be at least 12 characters long. However, you can customize the password length requirement:
 
 1. In the upper left corner, click **☰ > Global Settings**.
 1. Go to **`password-min-length`** and click **⋮ > Edit Setting**.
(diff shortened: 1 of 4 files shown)


sunilarjun (commented):
The instructions look good overall, but do we want to note in the v2.7 instructions that password validation was added to possible input as of v2.7.2? (Not totally sure it's needed, just wanted to suggest as the v2.6 phrasing states when password length customization was added.)

martyav:
@sunilarjun I'm kind of divided on it too...I think, ultimately, telling the user the requirements regardless of if the system actually validates the input provides guard rails for v2.6.9--2.7.1. If they enter bad input in those versions, the system won't reject it, but it's probably going to cause odd behavior.

sunilarjun:
> @sunilarjun I'm kind of divided on it too...I think, ultimately, telling the user the requirements regardless of if the system actually validates the input provides guard rails for v2.6.9--2.7.1. If they enter bad input in those versions, the system won't reject it, but it's probably going to cause odd behavior. So, telling the user what the input should be is more useful than telling them when a validation was introduced.

That makes sense and thanks for elaborating Marty. I'll go ahead and approve as I agree with the thought process.

sunilarjun (approved):

martyav:
I'm going to merge since I need this PR for the other password-related open PR (#1208)

PR #1218, opened 2024-04-05
Include documentation on how to make UI extensions work in airgap mode

martyav:
Fixes SURE-8077

## Reminders

- See the [README](../README.md) for more details on how to work with the Rancher docs.

- Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.

- If the pull request is dependent on an upcoming release, make sure to target the release branch instead of `main`.

## Description

Largely based on content from https://github.com/rancher/ui-plugin-charts?tab=readme-ov-file#air-gapped-rancher

Also reorganizes the air-gapped related content. Do you think maybe we should put it all on a separate page? Disadvantage is that some of other instructions are still relevant for air-gapped, so air-gapped users might not realize that deletion, uninstall, etc. are listed on another page.

PR #1222, opened 2024-04-08
Syncing with #1218 + rm'ing info about v2.7.0 from v2.8 page

martyav:
Related to #1218 

## Reminders

## Description

In addition to syncing the v2.8 page with latest, this removes some stray mentions of v2.7.

I'm only covering v2.8 because v2.8.2 is the associated version on the Jira, and only the v2.8 page currently mentions airgap instructions. I'm not sure if the instructions we list for airgapped are relevant to v2.7. The file itself only exists in v2.7 and v2.8 versioned docs.

### DIFF
--- a/docs/integrations-in-rancher/rancher-extensions.md
+++ b/docs/integrations-in-rancher/rancher-extensions.md
@@ -6,7 +6,7 @@ title: Rancher Extensions
   <link rel="canonical" href="https://ranchermanager.docs.rancher.com/integrations-in-rancher/rancher-extensions"/>
 </head>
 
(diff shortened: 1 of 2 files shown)


btat (approved):
> I'm not sure if the instructions we list for airgapped are relevant to v2.7. The file itself only exists in v2.7 and v2.8 versioned docs.

We should get confirmation since 2.7 is an active version and extensions was the main feature for it. None of the updates look 2.8 specific though.

Adding approval for current changes to /docs, and /version-2.8 to not block proceeding if /version-2.7 isn't needed.

PR #1235, opened 2024-04-15
Fix name of option being used on doc

Valgueiro:
## Description

Changed the description of the command being run to describe it correctly

## Comments

N/A


martyav (approved):
Looks like a typo slipped through. Thanks for correcting. We have to sync changes to the files in `versioned_docs` as well, but I can merge this and create a separate PR for that.

PR #1242, opened 2024-04-17
#1182 add warning note for node driver deletion on vmware

### LINKED ISSUE #1182: Add Warning Note for Node Driver Deletion on VMware
## Related Issues

After a customer disabled the VMware node driver, they faced serious challenges with their Rancher-managed nodes. The initial node failed, leading to the inaccessibility of all nodes in the cluster. Attempts to delete the nodes were also unsuccessful, as they would remain in an extended 'deleting' state. This experience highlighted a gap in the Rancher documentation, particularly on the Node Driver page, regarding the potential consequences of disabling or deleting a node driver, especially for critical drivers like VMware's.

## Summary

I propose adding a note or warning on the Node Driver page within the Rancher documentation to inform users about the critical impact of disabling or deleting a node driver. Specifically, it should highlight that "deleting a node driver will break clusters deployed via said driver."

## Details

- The warning should clearly state the consequences of disabling and deleting node drivers, using the VMware driver as a specific example.
- It could include a recommendation or steps on what to consider or check before taking such actions to avoid unintended disruptions.
- A link to troubleshooting steps or how to recover from such a state if it accidentally occurs would also be beneficial for users.
- This warning should be placed prominently on the Node Driver page, perhaps in a highlighted box or a separate section that easily catches the reader's attention.

martyav:
Fixes #1182 

## Reminders

This adds a warning about how deactivating node drivers for UI presentation reasons can have serious effects for already-deployed clusters.

### DIFF
--- a/docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/about-provisioning-drivers/manage-node-drivers.md
+++ b/docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/about-provisioning-drivers/manage-node-drivers.md
@@ -31,9 +31,9 @@ By default, Rancher only activates drivers for the most popular cloud providers,
 
 2.	On the **Node Drivers** tab, select the driver that you wish to activate or deactivate and click **⋮ > Activate** or **⋮ > Deactivate**.
 
-### Consequences of Deactivating or Deleting Node Drivers
+### Consequences of Deactivating Node Drivers
 
-When you deactivate or delete a node driver, the nodes deployed with that driver become inaccessible. For example, if you are have a vSphere cluster, and you disable the vSphere node driver used to deploy the cluster's nodes, the initial node in the cluster will fail, and the entire cluster will become inaccessible. Attempts to delete the VMware nodes will fail, with nodes stuck in an extended `Deleting` state.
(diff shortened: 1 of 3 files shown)


docs/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/about-provisioning-drivers/manage-node-drivers.md
```diff
@@ -21,17 +21,31 @@ To create, edit, or delete drivers, you need _one_ of the following permissions:
...
-## Adding Custom Node Drivers
+Deactivating a node driver doesn't just affect its visibility in the Rancher UI. When you deactivate or delete a node driver, any nodes deployed with that driver become inaccessible. 
+
+For example, if you deactivate a vSphere node driver to hide it in the UI, and you have a vSphere cluster that was deployed with that driver, the initial node in the cluster will fail, and the entire cluster will become inaccessible. Attempts to delete the vSphere nodes will fail, with nodes stuck in an extended `Removing` state.
+
+Before you deactivate a node driver, make sure that it has no associated clusters. One way to check is to see if the respective platform for a driver is listed among your clusters:
+
+1. In the upper left corner, click **☰ > Cluster Management**.
+1. Select **Clusters**.
+1. The **Provider** name is the fourth column of the table on the **Clusters** page.
```
btat, line 44: I think we should drop the column number to avoid it being incorrect if there are any layout changes in the future.

```suggestion
1. Check the **Provider** column of the table for instances of the node driver you are deactivating.
```

PR #1254, opened 2024-04-26
Update publish-images.md

super-small-rookie:
Add project creation instructions when pushing images and modify step numbers.

Fixes #[issue_number]

## Reminders

### DIFF
--- a/docs/getting-started/installation-and-upgrade/other-installation-methods/air-gapped-helm-cli-install/publish-images.md
+++ b/docs/getting-started/installation-and-upgrade/other-installation-methods/air-gapped-helm-cli-install/publish-images.md
@@ -283,8 +283,8 @@ Move the images in the `rancher-images.tar.gz` to your private registry using th
 
 The image list, `rancher-images.txt` or `rancher-windows-images.txt`, is expected to be on the workstation in the same directory that you are running the `rancher-load-images.sh` script. The `rancher-images.tar.gz` should also be in the same directory.
 
-1. Ensure that the rancher project is created in your private registry.
-   Otherwise, an error will be reported like project rancher not found.
+1. Ensure that the Rancher project is in your private registry.
+   If not, you'll see an error message, such as, "project rancher not found".
    
 2. Log into your private registry if required:
    ```plain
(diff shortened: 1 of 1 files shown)


docs/getting-started/installation-and-upgrade/other-installation-methods/air-gapped-helm-cli-install/publish-images.md
```diff
@@ -283,17 +283,20 @@ Move the images in the `rancher-images.tar.gz` to your private registry using th
 
 The image list, `rancher-images.txt` or `rancher-windows-images.txt`, is expected to be on the workstation in the same directory that you are running the `rancher-load-images.sh` script. The `rancher-images.tar.gz` should also be in the same directory.
 
-1. Log into your private registry if required:
+1. Ensure that the rancher project is created in your private registry.
+   Otherwise, an error will be reported like project rancher not found.
+   
+2. Log into your private registry if required:
```
martyav, line 289:
```suggestion
1. Ensure that the Rancher project is in your private registry.
   If not, you'll see an error message, such as, "project rancher not found".
   
2. Log into your private registry if required:
```

martyav (changes requested):
Small copyedit.

btat (changes requested):
Can you add more context around your setup and at what step the error is being thrown? Is the concept of a project something specific to your private registry implementation?

martyav:
@super-small-rookie Hi, not sure if you saw the comment above. Could you answer these questions?

> Can you add more context around your setup and at what step the error is being thrown? Is the concept of a project something specific to your private registry implementation?

super-small-rookie:
I use the harbor registry, because the script need project which named rancher to push images. The rancher project does not exist in harbor by default, so it return information about project rancher not found. The information may be returned by harbor. I create the project rancher in harbor, then it pushed images well.

btat:
In the [preceding step where the docs talk about setting up a private registry](https://github.com/rancher/rancher-docs/blob/e8185388581e4354847553b0ce0a06c1592cb77e/docs/getting-started/installation-and-upgrade/other-installation-methods/air-gapped-helm-cli-install/infrastructure-private-registry.md) there are no references to any specific private registry implementations so changing this subsequent step to assume Harbor is used does not make sense.

PR #1280, opened 2024-05-09
Updating vale.yml - continue-on-error

sunilarjun:
Updating the vale.yml file with continue-on-error: true for the errata-ai/vale-action to succeeed even if the specified action fails. This is to have the vale style checker act as more of a warning and have less friction for contributors.

Test with [previously failing runner](https://github.com/rancher/rancher-docs/actions/runs/8993407297/job/24705151919) due to number of annotations that now passes with the `continue-on-error` addition:
https://github.com/rancher/rancher-docs/actions/runs/9019545283/job/24782655577?pr=1147

### DIFF
--- a/.github/workflows/vale.yml
+++ b/.github/workflows/vale.yml
@@ -45,6 +45,7 @@ jobs:
             echo ${ALL_CHANGED_FILES}
             echo "ALL_CHANGED_FILES=$ALL_CHANGED_FILES" >> $GITHUB_ENV
       - uses: errata-ai/vale-action@v2.1.0
+        continue-on-error: true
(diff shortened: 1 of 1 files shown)


btat (changes requested):
Can we add `continue-on-error` to the other steps too so that we can cover cases like https://github.com/rancher/rancher-docs/actions/runs/9005576107/job/24741210013?pr=827 where the failure is elsewhere.

sunilarjun:
Just applied to all steps and tested and got a [passing runner](https://github.com/rancher/rancher-docs/actions/runs/9020470293/job/24785693855?pr=1147), thanks for the review @btat !

btat (approved):

PR #1321, opened 2024-06-05
Update Versions Table - v2.8.5.

sunilarjun:
Updating the versions table fields for v2.8.5.


martyav (approved):
You'll be pleased to know that AsciiDoc supports text alignment for tables, so we can center things without using tags

PR #1401, opened 2024-07-24
Correct the default amount of time between Rancher cache syncs

moio:
## Description

Fixes a "typo" in the value of a Rancher default value which was documented incorrectly since the beginning. For some reason nobody noticed so far (😢).

Ultimate source of truth is 600 minutes = 10 hours per this code reference:

https://github.com/rancher/lasso/blob/32d96874d344eeeed641df40bdd38c50ad0213e7/pkg/cache/cache.go#L19-L20


martyav (approved):

martyav:
Thank you for correcting this!

PR #1422, opened 2024-08-02
Configure dependabot for vale rules

btat:
## Description

Set up dependabot so we don't need to manually track when new changes are made upstream.


martyav (approved):
Thank you for updating this

PR #1425, opened 2024-08-08
Use crds.enabled to install cert-manager

dharmit:
Fixes #[issue_number] N/A

## Description

Installing cert-manager using the currently documented command shows the following warning:
```
NOTES:
⚠️  WARNING: `installCRDs` is deprecated, use `crds.enabled` instead.
```

### DIFF
--- a/docs/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/install-upgrade-on-a-kubernetes-cluster.md
+++ b/docs/getting-started/installation-and-upgrade/install-upgrade-on-a-kubernetes-cluster/install-upgrade-on-a-kubernetes-cluster.md
@@ -148,7 +148,7 @@ To see options on how to customize the cert-manager install (including for cases
 :::
 
 ```
-# If you have installed the CRDs manually instead of with the `--set installCRDs=true` option added to your Helm install command, you should upgrade your CRD resources before upgrading the Helm chart:
+# If you have installed the CRDs manually instead of with the `--set crds.enabled=true` option added to your Helm install command, you should upgrade your CRD resources before upgrading the Helm chart:
 kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/<VERSION>/cert-manager.crds.yaml
 
 # Add the Jetstack Helm repository
@@ -161,7 +161,7 @@ helm repo update
 helm install cert-manager jetstack/cert-manager \
   --namespace cert-manager \
   --create-namespace \
-  --set installCRDs=true
+  --set crds.enabled=true
 ```
 
 Once you’ve installed cert-manager, you can verify it is deployed correctly by checking the cert-manager namespace for running pods:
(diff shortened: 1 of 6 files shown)


martyav:
@dharmit do you happen to know if this affects multiple Rancher versions? If just one, which one?

btat:
> @dharmit do you happen to know if this affects multiple Rancher versions? If just one, which one?

This is coming from cert-manager charts >= v1.15.0 and has no dependency on Rancher itself. This will come up for every cert-manager installation example we have where we don't explicitly set a version (that's also < v1.15.0) in which case it defaults to the latest version of the chart. It's safe to assume all instances of our examples fall into this category.

martyav:
@btat @LucasSaintarbor @sunilarjun could you have a look? I can't approve as I was the last to update the branch

sunilarjun (commented):
Should we update the translation versions in this PR or create another to update those pages? Otherwise, looks good.

btat (changes requested):
@martyav As pointed out in [this comment](https://github.com/rancher/rancher-docs/pull/1425#issuecomment-2276445648), this only affects cert-manager >= 1.15.0 (which is when `crds.enabled` was introduced). This means the blanket change in 0c56e643ad9dc4bb78939312bc0e3b3d9e6d506a introduces errors since some of our older docs are explicitly installing versions older than 1.15.0.

As for the cases where `crds.enabled` does apply, simply replacing `installCRDs` in the comment line without any context can be confusing. If a user previously used the old value and reads this page it may look like their case is not covered.

dharmit:
> Should we update the translation versions in this PR or create another to update those pages? Otherwise, looks good.

@sunilarjun I don't know how to add docs for language other than English. I'll need someone's help with that.



> @martyav As pointed out in [this comment](https://github.com/rancher/rancher-docs/pull/1425#issuecomment-2276445648), this only affects cert-manager >= 1.15.0 (which is when `crds.enabled` was introduced). This means the blanket change in [0c56e64](https://github.com/rancher/rancher-docs/commit/0c56e643ad9dc4bb78939312bc0e3b3d9e6d506a) introduces errors since some of our older docs are explicitly installing versions older than 1.15.0.
> 
> As for the cases where `crds.enabled` does apply, simply replacing `installCRDs` in the comment line without any context can be confusing. If a user previously used the old value and reads this page it may look like their case is not covered.

@btat I'm not sure if you are suggesting me to make some changes here.

btat:
> > Should we update the translation versions in this PR or create another to update those pages? Otherwise, looks good.
> 
> @sunilarjun I don't know how to add docs for language other than English. I'll need someone's help with that.
> 
> > @martyav As pointed out in [this comment](https://github.com/rancher/rancher-docs/pull/1425#issuecomment-2276445648), this only affects cert-manager >= 1.15.0 (which is when `crds.enabled` was introduced). This means the blanket change in [0c56e64](https://github.com/rancher/rancher-docs/commit/0c56e643ad9dc4bb78939312bc0e3b3d9e6d506a) introduces errors since some of our older docs are explicitly installing versions older than 1.15.0.
> > As for the cases where `crds.enabled` does apply, simply replacing `installCRDs` in the comment line without any context can be confusing. If a user previously used the old value and reads this page it may look like their case is not covered.
> 
> @btat I'm not sure if you are suggesting me to make some changes here.

@dharmit My comment was intended for @martyav since during the process of helping sync the update across different versions of our docs, it introduced errors into the older docs (<= 2.6). I believe Marty is already planning to address those, but feel free to make the changes yourself if preferred. As for the translation files, Marty can probably help with those as well. If you're curious, those files are at https://github.com/rancher/rancher-docs/tree/main/i18n/zh/docusaurus-plugin-content-docs.

btat (approved):

PR #1444, opened 2024-08-26
[2.9.1] Impersonation - Known Issue Section

sunilarjun:
Adding in known issue warning to Impersonation section for v2.9.1.

Tied to https://github.com/rancher/rancher/issues/46790

### DIFF
--- a/docs/reference-guides/rancher-manager-architecture/communicating-with-downstream-user-clusters.md
+++ b/docs/reference-guides/rancher-manager-architecture/communicating-with-downstream-user-clusters.md
@@ -89,6 +89,12 @@ We recommend exporting the kubeconfig file so that if Rancher goes down, you can
 
 ## Impersonation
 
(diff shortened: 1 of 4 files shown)


btat (approved):
Approving.

Small nit that shouldn't block PR: should 'Known Issues' be singular here? Comparing it to when we use the note admonition, we don't use the plural form.

sunilarjun:
> Approving.
> 
> Small nit that shouldn't block PR: should 'Known Issues' be singular here? Comparing it to when we use the note admonition, we don't use the plural form.

Agreed on it being singular, just edited. Thanks for the catch Billy!

martyav:
Why no update to 2.6 versioned docs? https://github.com/rancher/rancher/issues/41988#issuecomment-1790820817

sunilarjun:
I just made updates to maintained versions but I can add this update to v2.6 as well.

PR #1460, opened 2024-09-10
Fix text errors

btat:


btat:
Already being handled in #1459.

PR #1488, opened 2024-09-19
zh/2.8 fix header levels

btat:
## Description

Ensures section headers start at h2 and increment one level at a time.


sunilarjun (approved):
Thanks Billy!

PR #1509, opened 2024-10-16
Remove vale workflow - contains unapproved GHA

btat:
## Description

This PR disables the Vale style check workflow as it currently uses actions that aren't org-approved. To be re-enabled when an alternate implementation is made.

The [style rules](https://github.com/rancher/rancher-docs/tree/main/.github/styles) aren't removed since they can still be used for local verification with the Vale CLI.


sunilarjun (approved):
LGTM

PR #1526, opened 2024-10-25
Set max-old-space-size for yarn start due to memory issues

btat:
## Description

We've received feedback that using `yarn start` to preview the docs also encounters the `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` error that's observed during `yarn build`.

This PR applies the same [memory setting](https://github.com/rancher/rancher-docs/blob/main/package.json#L8) that we already have for `yarn build`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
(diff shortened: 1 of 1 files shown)


sunilarjun (approved):
Thanks Billy!

PR #1531, opened 2024-10-29
Improve encrypted backup docs

jbiers:
Fixes [#593](https://github.com/rancher/backup-restore-operator/issues/593)

### DIFF
--- a/docs/reference-guides/backup-restore-configuration/examples.md
+++ b/docs/reference-guides/backup-restore-configuration/examples.md
@@ -250,10 +250,26 @@ spec:
 
 The snippet below demonstrates two different types of secrets and their relevance with respect to Backup and Restore of custom resources.
 
-The first example is that of a secret that is used to encrypt the backup files. The backup operator, in this case, will not be able to read the secrets encryption file. It only uses the contents of the secret.
+The first example is that of a secret used to encrypt the backup files. The backup operator will read the contents of the **encryption-provider-config.yaml** key, which contains the definition of an EncryptionConfiguration resource encoded as Base64. Creating the secret can be done with the following command:
 
-The second example is that of a Kubernetes secrets encryption config file that is used to encrypt secrets when stored in etcd. **When backing up the etcd datastore, be sure to also back up the EncryptionConfiguration.** Failure to do so will result in an inability to use the restored data if secrets encryption was in use at the time the data was backed up.
+```plain
+kubectl create secret generic example-encryptionconfig \
+  --from-file=./encryption-provider-config.yaml \
+  -n cattle-resources-system
+```
 
+The second example is that of the Kubernetes EncryptionConfiguration file itself, which is also used to encrypt secrets when stored in etcd. **When backing up the etcd datastore, be sure to also back up the EncryptionConfiguration.** Failure to do so will result in an inability to use the restored data if secrets encryption was in use at the time the data was backed up.
+
+```yaml
+apiVersion: v1
+data:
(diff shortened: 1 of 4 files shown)


docs/reference-guides/backup-restore-configuration/examples.md
```diff
@@ -250,17 +250,34 @@ spec:
 
 The snippet below demonstrates two different types of secrets and their relevance with respect to Backup and Restore of custom resources.
```
mallardduck, line 251: I think it might be better "flow" if the part about creating the secret was here.
So both "Creating this secret can be done with the following command:" and the example command you added here.

docs/reference-guides/backup-restore-configuration/examples.md
```diff
@@ -250,17 +250,34 @@ spec:
...
 
-The second example is that of a Kubernetes secrets encryption config file that is used to encrypt secrets when stored in etcd. **When backing up the etcd datastore, be sure to also back up the EncryptionConfiguration.** Failure to do so will result in an inability to use the restored data if secrets encryption was in use at the time the data was backed up.
+```plain
+kubectl create secret generic example-encryptionconfig \
+  --from-file=./encryption-provider-config.yaml \
+  -n cattle-resources-system
+```
+
+The second example is that of the Kubernetes EncryptionConfiguration file itself, which is also used to encrypt secrets when stored in etcd. **When backing up the etcd datastore, be sure to also back up the EncryptionConfiguration.** Failure to do so will result in an inability to use the restored data if secrets encryption was in use at the time the data was backed up.
```
mallardduck, line 261: Then down here it will read as 2 paragraphs each talking about the respective example to follow.

docs/reference-guides/backup-restore-configuration/examples.md
```diff
@@ -250,17 +250,34 @@ spec:
...
 
+```yaml
+apiVersion: v1
+data:
+  encryption-provider-config.yaml: YXBpVmVyc2lvbjogYXBpc2VydmVyLmNvbmZpZy5rOHMuaW8vdjEKa2luZDogRW5jcnlwdGlvbkNvbmZpZ3VyYXRpb24KcmVzb3VyY2VzOgogIC0gcmVzb3VyY2VzOgogICAgICAtIHNlY3JldHMKICAgICAgLSAqLmFwcHMKICAgIHByb3ZpZGVyczoKICAgICAgLSBhZXNnY206CiAgICAgICAgICBrZXlzOgogICAgICAgICAgICAtIG5hbWU6IGtleTEKICAgICAgICAgICAgICBzZWNyZXQ6IGMyVmpjbVYwSUdseklITmxZM1Z5WlE9PQogICAgICAgICAgICAtIG5hbWU6IGtleTIKICAgICAgICAgICAgICBzZWNyZXQ6IGRHaHBjeUJwY3lCd1lYTnpkMjl5WkE9PQogICAgICAtIGFlc2NiYzoKICAgICAgICAgIGtleXM6CiAgICAgICAgICAgIC0gbmFtZToga2V5MQogICAgICAgICAgICAgIHNlY3JldDogYzJWamNtVjBJR2x6SUhObFkzVnlaUT09CiAgICAgICAgICAgIC0gbmFtZToga2V5MgogICAgICAgICAgICAgIHNlY3JldDogZEdocGN5QnBjeUJ3WVhOemQyOXlaQT09CiAgICAgIC0gc2VjcmV0Ym94OgogICAgICAgICAga2V5czoKICAgICAgICAgICAgLSBuYW1lOiBrZXkxCiAgICAgICAgICAgICAgc2VjcmV0OiBZV0pqWkdWbVoyaHBhbXRzYlc1dmNIRnljM1IxZG5kNGVYb3hNak0wTlRZPQo=
```
mallardduck, line 266: I think that this one could _technically_ be OK to be plain-text. Since showing it as base64 obfuscates the compare/contrast that I believe the section here is trying to communicate.

Since users should be learning (by example) that the Etcd Encryption config and the Backup Restore Operator config are the same idea, but plugged into different places for different contexts.

docs/reference-guides/backup-restore-configuration/examples.md
```diff
@@ -250,17 +250,34 @@ spec:
...
+type: Opaque
+```
 
 ```yaml
 apiVersion: apiserver.config.k8s.io/v1
 kind: EncryptionConfiguration
 resources:
   - resources:
       - secrets
+      - *.apps
```
mallardduck, line 280: I would leave it to a simple example of just secrets. 😸

mallardduck (changes requested):
Great start overall, just a few tweaks I had in mind that may improve the readability flow.

mallardduck (approved):
👍  LGTM

btat (changes requested):
Content changes LGTM, but adding review as 'request changes' since the updates need to be ported to other versions of our docs. Let me know if assistance is needed with that.

PR #1567, opened 2024-11-14
Remove mentions of rancher telemetry

adamkpickering:
In https://github.com/rancher/rancher/pull/48031 we removed any use of https://github.com/rancher/telemetry, and thus telemetry in general, in Rancher Manager. This PR removes mentions of telemetry in the docs.

Please give this a thorough review since I'm not super familiar with this repository. My understanding is that the telemetry removal is planned for 2.11, so I think that we only need to worry about files in `docs/`.


l-mb (approved):
Looks good to me, thanks.

sunilarjun (approved):
Looks good, thanks Adam!

PR #1576, opened 2024-11-20
Remove extra image captions

btat:
## Description

The image on `high-availability-installs.md` has 2 captions with the same text. One with `<figcaption>`, but the resulting rendering is indistinguishable from regular paragraph text. The other uses `<sup>` to try and emulate an actual caption. I opted to remove the `<figcaption>` instance due to the lack of formatting.

On `how-monitoring-works.md`, there's a figcaption instance that has no image surrounding it.


sunilarjun (approved):
LGTM

PR #1589, opened 2024-12-10
Syncing CVE Pages Across Docs/i18n Versions

sunilarjun:
Syncing v2.10 CVE page as some CVE's were missing from previous updates.


btat (changes requested):
The starting pipe character is missing from the start of each row.

Note: also seems to be the case for [original PR](https://github.com/rancher/rancher-docs/pull/1583) that's referenced.

sunilarjun:
> The starting pipe character is missing from the start of each row.
> 
> Note: also seems to be the case for [original PR](https://github.com/rancher/rancher-docs/pull/1583) that's referenced.

@btat Thanks for catching Billy! I've pushed a fix to address the missing pipe character between versions, and also realized that the i18n CVE versions were out of sync for a few updates so added those in with this PR as well. Let me know if anything needs to be adjusted, thank you!

btat (dismissed):
Thanks @sunilarjun LGTM. One small nit is that 22030 could have its translation preserved, but adding an approval since it's not a blocker.

sunilarjun:
@btat Thanks for catching that! Added the translated version back in and added back an incorrectly removed CVE from v2.7, please let me know if anything else needs editing.

PR #1594, opened 2024-12-16
chore: improve readability

nhecker:
## Description
This replaces instances of a word that is easy enough to parse for a native English speaker, but isn't really a word and may be harder to understand than an alternate phrasing. There are no other GH artifacts associated with this PR.

## Comments
This is just a quickie edit in the GH WebUI, so the commit isn't signed, etc. If you need that just let me know and I'll resubmit the PR. Thanks!


btat (approved):

LucasSaintarbor (approved):

btat:
Thank you for the contribution!

PR #1602, opened 2024-12-20
Removing Incorrect Directories - [SURE-8597]

sunilarjun:
Removing incorrect directories from list of directories used by RKE1, RKE2, and K3s. Also updating admonition.
Tied to SURE-8597

### DIFF
--- a/docs/how-to-guides/new-user-guides/manage-clusters/clean-cluster-nodes.md
+++ b/docs/how-to-guides/new-user-guides/manage-clusters/clean-cluster-nodes.md
@@ -236,10 +236,8 @@ sudo k3s-uninstall.sh
 
 The following directories are used when adding a node to a cluster, and should be removed. You can remove a directory using `rm -rf /directory_name`.
 
-:::note
-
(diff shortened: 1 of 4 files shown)


btat (changes requested):
In the case of `/var/lib/etcd`, it may be valid for RKE. A quick search shows that there are instances of it in the code bases for [rancher/rke](https://github.com/search?q=repo%3Arancher%2Frke%20%2Fvar%2Flib%2Fetcd&type=code) and [rancher/rancher](https://github.com/search?q=repo%3Arancher%2Francher%20%2Fvar%2Flib%2Fetcd&type=code), but I don't know if those bits of code correlate to this ticket.

Also, the feedback from the ticket sounds like it's based off RKE2 so just removing the specific examples mentioned may not be the appropriate solution since the table represents all three distros right now. Having separate lists might help with that. E.g. `opt/rke` is not mentioned in the ticket, but it's not tied to K3s.

> Update documentation with correct list of directories used by rke1, rke2, and k3s

It's not mentioned in the PR description, was the rest of the list verified?

Issue #1176, opened 2024-03-14
Improve OIDC / Rancher/Keycloak docs


LucasSaintarbor:
## Related Issues
SURE-6968

## Summary

RGS has the below docs request that seems to align with 

https://github.com/rancher/rancher-docs/pull/827

- We want to investigate the possibility of incorporating the following changes into the Rancher/Keycloak documentation: 
To enable the ability to view both groups and users, simply add the realm-management based role ‘view-users’ Adding the client role ‘view-users’ (under realm management) at either the group level or individual user level will enable the ability to search for users in Rancher.  https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-keycloak-oidc 
- Customers have encountered the following problem concerning the Keycloak OIDC configuration for users and groups. After adding new groups in Keycloak, these groups are not reflected in Rancher. Customers cannot search for new groups or users or manually input them via the ‘Authorized users and groups’ section. Having the view-users client scope allowed us to see the users/groups as desired.

Issue #1114, opened 2024-02-09
Keycloak related docs are outdated


martyav:
## Summary

From Slack (@tlatino):

> Has [https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permiss[…]uration/authentication-config/configure-keycloak-oidc](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-keycloak-oidc) been looked into for newer keycloak versions? It seems to be out of date. I am walking a customer through the process and they are finding that most of the suggestions are out of date for more recent versions of keycloak causing confusion and frustration.

tlatino-suse:
For lower privileged users in Keycloak, user/group lookup may not work as expected unless the user is assigned "view-users" in Keycloak under the role mapping section of the user for the appropriate realm. Once assigned, they can search for Keycloak users and groups in Rancher by starting to type the usernames in the search bar. This is not pointed out in the docs and should be, as if a user is not assigned "view-users", they can not assign permissions to any keycloak user in Rancher because they will not be able to search for them.

Expand on how to add mappers in keycloak because the keycloak documentation does not make this clear. In the newer Keycloak docs, the user will need to click on the client page, click on the client they had created, click on "client scopes", click on the assigned client scope by the name "<CLIENT_ID>-dedicated", click "Add Mapper", click "By Configuration", and select the "Mapper Type" in the tables found in https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/authentication-permissions-and-global-configuration/authentication-config/configure-keycloak-oidc

tlatino-suse:
Expanding on this, the "view-users" role is necessary if users need the permissions to search through all groups that exist in Keycloak under their realm as well. If they do not have the "view-users" role in Keycloak, they can only assign roles to groups they themselves are a part of.

PR #1641, opened 2025-02-12
Update field names and nav paths to reflect changes in newer Keycloak…

btat:
SURE-9548

Fixes https://github.com/rancher/rancher-docs/issues/1114, fixes https://github.com/rancher/rancher-docs/issues/1176

## Description

Keycloak, including its docs, has undergone many revisions (e.g. new UI, migrated to Quarkus), which has caused a disconnect between what's indicated mentioned in our docs and what's in recent versions of Keycloak.

This PR updates references fields and navigation paths to sync with current versions (25/26) of Keycloak. Our existing docs list certain fields and their expected values, but newer Keycloak versions have additional fields (e.g. `lightweight access token` and `Add to token introspection` for the Group membership mapper type). Per engineering, these aren't used and can be left with the default value.

I also opted to list out explicit Keycloak steps as I either couldn't find any relevant links or if linkable steps were present it needed additional context as the relevant steps were interwoven with other content.


sunilarjun (approved):
Smoke tested per version/translation and the changes look good, thanks for updating Billy! Approved

PR #1216, opened 2024-04-04
Adding Rancher Turtles Overview - Integrations with Rancher Section

sunilarjun:
Adding the CAPI integrations dropdown and overview page to Rancher latest, v2.8, and v2.7. Adjusting the sidebars for the respective versions as well. The [CAPI documentation](https://docs.rancher-turtles.com/) site was used for reference. The overview page consists of installing/uninstalling Rancher Turtles, an architecture diagram, and extension security.

PR #1642, opened 2025-02-13
Include CAPI as a tile in integrations page

btat:
## Description

When we initially add the CAPI/Turtles docs in #1216, we forgot to adjust its order and update the card list array range on the integrations index page. As a result, https://ranchermanager.docs.rancher.com/integrations-in-rancher does not show CAPI as a tile.


LucasSaintarbor (commented):
LGTM. But there is a broken link for [CAPI Operator](https://turtles.docs.rancher.com/reference-guides/rancher-turtles-chart/values#cluster-api-operator-values) (second bullet point on the Cluster API (CAPI) with Rancher Turtles page. I think it should be: https://turtles.docs.rancher.com/turtles/next/en/reference-guides/rancher-turtles-chart/values.html#_cluster_api_operator_values. Not sure if it should be fixed with this PR.

btat:
> LGTM. But there is a broken link for [CAPI Operator](https://turtles.docs.rancher.com/reference-guides/rancher-turtles-chart/values#cluster-api-operator-values) (second bullet point on the [Cluster API (CAPI) with Rancher Turtles](http://localhost:3000/integrations-in-rancher/cluster-api) page). I think it should be: https://turtles.docs.rancher.com/turtles/next/en/reference-guides/rancher-turtles-chart/values.html#_cluster_api_operator_values. Not sure if it should be fixed with this PR.

I will fix that separately to not scope creep since it's not directly tied to the changes here and there are multiple link breakages like that one because of the Docusaurus to Antora change on the CAPI site.

LucasSaintarbor (approved):

PR #1660, opened 2025-02-25
Add issue template for release maintenance checklist

btat:
## Description

We should take advantage of templates since It's a recurring issue that we create for each release.

Also cleaned up some optional fields that we specified, but didn't use in our other templates.


sunilarjun (approved):
Thanks Billy!

LucasSaintarbor (approved):

Issue #1632, opened 2025-02-04
users should handle policy exemptions


samjustus:
## Summary

The policies shipped by default in Rancher aim to provide a trade-off between security and convenience. If users aim for a more strict policy, they should craft such policies themselves, based on their specific requirements. In case they prefer relying on Rancher policies, they will need to deploy admission controllers that block the creation of any exempted namespaces that won't be used within their environments.

btat:
Addressed by https://github.com/rancher/rancher-docs/pull/1704.

PR #1704, opened 2025-03-11
[v2.11] Policy Exemption User Note

sunilarjun:
Adding note to PSA template on users handling policy exemptions, tied to Rancher docs issue [1632](https://github.com/rancher/rancher-docs/issues/1632).


sunilarjun:
@btat Thanks for reviewing Billy! I don't think it's needed as I took the copy from Sam's issue so should be good to merge.

PR #1705, opened 2025-03-12
backup and restore: add warning not to use snapshots

moio:
## Description

Relying on etcd snapshots for the upstream ("local") cluster is known to cause issues.

The Backup and Restore Operator was created to avoid those.

We should clarify that to customers.

## Comments

I opened a PR instead of an issue in hopes that will help fixing the problem sooner - and also because it was the easiest way for me to communicate clearly what the problem is. I hope this helps

### DIFF
--- a/docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
+++ b/docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
@@ -10,6 +10,14 @@ The Rancher Backups chart is our solution for disaster recovery and migration. T
 
 This chart is a very simple tool which has its hands in many different areas of the Rancher ecosystem. As a result, edge cases have arisen which lead to undocumented functionality. The purpose of this document is to highlight the proper and defined usage for Rancher Backups, as well as discussing some of these edge cases we’ve run into.
 
+:::caution
+
+While Kubernetes distributions like k3s and RKE2 offer [cluster configuration backups](https://docs.rke2.io/datastore/backup_restore) (etcd snapshots), these can sometimes cause issues with Rancher.
+
+For a reliable Rancher backup, we recommend using the Rancher Backup functionality described in this guide. Consider etcd snapshots as a last resort option.
+
+:::
+
 ## Functionality Overview
 
 ### Backup
--- a/versioned_docs/version-2.10/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
+++ b/versioned_docs/version-2.10/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
@@ -10,6 +10,14 @@ The Rancher Backups chart is our solution for disaster recovery and migration. T
 
 This chart is a very simple tool which has its hands in many different areas of the Rancher ecosystem. As a result, edge cases have arisen which lead to undocumented functionality. The purpose of this document is to highlight the proper and defined usage for Rancher Backups, as well as discussing some of these edge cases we’ve run into.
 
+:::caution
+
+While Kubernetes distributions like k3s and RKE2 offer [cluster configuration backups](https://docs.rke2.io/datastore/backup_restore) (etcd snapshots), these can sometimes cause issues with Rancher.
+
+For a reliable Rancher backup, we recommend using the Rancher Backup functionality described in this guide. Consider etcd snapshots as a last resort option.
+
+:::
+
 ## Functionality Overview
 
 ### Backup
--- a/versioned_docs/version-2.11/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
+++ b/versioned_docs/version-2.11/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
@@ -10,6 +10,14 @@ The Rancher Backups chart is our solution for disaster recovery and migration. T
 
 This chart is a very simple tool which has its hands in many different areas of the Rancher ecosystem. As a result, edge cases have arisen which lead to undocumented functionality. The purpose of this document is to highlight the proper and defined usage for Rancher Backups, as well as discussing some of these edge cases we’ve run into.
 
+:::caution
+
+While Kubernetes distributions like k3s and RKE2 offer [cluster configuration backups](https://docs.rke2.io/datastore/backup_restore) (etcd snapshots), these can sometimes cause issues with Rancher.
+
+For a reliable Rancher backup, we recommend using the Rancher Backup functionality described in this guide. Consider etcd snapshots as a last resort option.
+
+:::
+
 ## Functionality Overview
 
 ### Backup
--- a/versioned_docs/version-2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
+++ b/versioned_docs/version-2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
@@ -10,6 +10,14 @@ The Rancher Backups chart is our solution for disaster recovery and migration. T
 
 This chart is a very simple tool which has its hands in many different areas of the Rancher ecosystem. As a result, edge cases have arisen which lead to undocumented functionality. The purpose of this document is to highlight the proper and defined usage for Rancher Backups, as well as discussing some of these edge cases we’ve run into.
 
+:::caution
+
+While Kubernetes distributions like k3s and RKE2 offer [cluster configuration backups](https://docs.rke2.io/datastore/backup_restore) (etcd snapshots), these can sometimes cause issues with Rancher.
+
+For a reliable Rancher backup, we recommend using the Rancher Backup functionality described in this guide. Consider etcd snapshots as a last resort option.
+
+:::
+
 ## Functionality Overview
 
 ### Backup
--- a/versioned_docs/version-2.7/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
+++ b/versioned_docs/version-2.7/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
@@ -10,6 +10,14 @@ The Rancher Backups chart is our solution for disaster recovery and migration. T
 
 This chart is a very simple tool which has its hands in many different areas of the Rancher ecosystem. As a result, edge cases have arisen which lead to undocumented functionality. The purpose of this document is to highlight the proper and defined usage for Rancher Backups, as well as discussing some of these edge cases we’ve run into.
 
+:::caution
+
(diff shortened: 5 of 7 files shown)


moio:
cc @mallardduck please take a look if this makes sense from your perspective
cc @jakefhyde please take a look if this makes sense from your perspective
cc @kourosh7 for initially bringing this topic up

mallardduck:
From my understanding the limitations of etcd as an option for local is more focused on RKE1 compared to other runtimes. For instances I know some PSEs that have customers that _only_ use Etcd for local backups and do not use BRO at all.

So I think we should make sure we capture what that nuance is to properly communicate that. As from my understanding ETCD is a safe option for full local (rancher) cluster backups when using k3s/rke2 for instance.

jakefhyde:
> From my understanding the limitations of etcd as an option for local is more focused on RKE1 compared to other runtimes. For instances I know some PSEs (@inichols might be able to give some input) that have customers that _only_ use Etcd for local backups and do not use BRO at all.
> 
> So I think we should make sure we capture what that nuance is to properly communicate that. As from my understanding ETCD is a safe option for full local (rancher) cluster backups when using k3s/rke2 for instance.
> 
> From the BRO docs, we specifically call out that BRO/Rancher Backups is not an etcd replacement. So we should be sure to not create a conflicting message between these docs and our [BRO readme](https://github.com/rancher/backup-restore-operator?tab=readme-ov-file#use-cases).
> 
> > What the Backup Restore Operator is not:
> > 
> > * A downstream cluster snapshot tool,
> > * A replacement for Etcd cluster backups,
> > * Configured to back up user-created resources on the Rancher cluster.

@mallardduck Want to clear up some things:
- ETCD Snapshot Pros
  - Easy, often automatic
  - Useful if nothing was changed since the last snapshot
- ETCD Snapshot Cons
  - If anything was changed in the local cluster, those changes are lost (i.e. CAPI resources)
    - If all of the machines were rotated for a cluster after the snapshot, those machines are "orphaned" and the cluster will also need to be restored with manual intervention
    - If clusters were created or deleted after the snapshot was taken, they will be orphaned or recreated respectively

Though simpler, the cons are very bad here. Since the state of downstreams is stored within the local cluster, I wouldn't recommend users to rely only on snapshots if they are performing massive infrastructure changes. I think this PR is correct to say that etcd restores should be treated as a last resort, because BRO has the advantage of being able to restore a local cluster with a partial state. We have users that rotate the nodes in their clusters as part of maintenance, so having a selective restoration tool that allows you to rollback to a certain state without losing everything created since makes resolving issues much simpler. I don't think any of that contradicts what is in https://github.com/rancher/backup-restore-operator?tab=readme-ov-file#use-cases, since it's filling a use-case which etcd snapshots are not suitable for in the first place.

jakefhyde:
> What issues, when?

@mallardduck I think that's a reasonable criticism. I think we would do well to inform the user about when one is recommended over the other, and what the potential tradeoffs are. IMO etcd restore would even be usable for rancher version rollbacks, as long as the upgrade didn't cause any changes to downstream clusters, but we never really guarantee that (hard to anyway, given the cluster-agent will immediately be rolled out on upgrade).

mallardduck:
@jakefhyde - I wonder between your feedback and mine, if a good way to communicate all this would be a new page specific to comparing and contrasting Etcd (specifically in context to local cluster only) vs Rancher backup? Then the snippet we add to callout this aspect on this page could read more like:

> For a reliable Rancher backup, we recommend using the Rancher Backup functionality described in this guide. Read our "(insert title)" guide to understand our best practices on when to use Rancher Backups or etcd backups.

Then we can include some of these details that better explain the risk we called out in the paragraph above this line? @moio What do you think of these additions to add more clarity around the two tools - and providing a guide to compare/contrast so users can pick the best strategy (a mix of both hopefully) for their use-cases?

moio:
I think adding a page with a more detailed analysis, and link it from the warning, is the best idea.

My problem is I do not feel like being competent enough to be authoritative about the details. Best I can do is to copypaste what you mentioned above.

Can any of you draft such a page? You have push permissions to this branch, just push a new markdown file in and let's iterate on the content.

Then I can take care of the fine editing, rancher-docs specifics, and improving on the Doc Team feedback until the PR is merged.

jakefhyde:
@moio I can draft something up tomorrow

moio:
Thank you.

BTW I consider this an "important not urgent" topic, so it's perfectly fine waiting a week or two as well, as long as it isn't forgotten 😇

jakefhyde:
@moio This is what I have come up with, feel free to tweak (also @mallardduck feel free to critique as well).
```
In the evevent of a failure in the local cluster, it may be necessary to restore Rancher to a previous version. There exist two currently supported methods for restoring Rancher: ETCD Restores, and restores performed by Rancher's Backup Restore Operator. ETCD Restores are suitable if no user changes have been made since the snapshot was taken which need to be persisted. For example, if a user takes a snapshot and then makes a change to their local cluster which causes an outage, an ETCD Restore would be the expected restoration mechanism. However, if a user takes a snapshot of the local cluster and then upgrades the kubernetesVersion of the downstream cluster, an ETCD restore would reset the kubernetesVersion and the nodes in this cluster would be running a version of kubernetes that is higher than their spec. Rancher will atempt to reconcile the version, which could have disastrous results. Likewise, if a user were to take a snapshot of the local cluster, rotate all of the nodes in a Rancher provisioned downstream cluster, and then restore from the snapshot, the information on how these nodes were provisioned would be lost, these nodes would be orphaned and in some cases the cluster could become inoperable. As a general rule, when restoring Rancher if any meaningful changes have been made to the local cluster (installing charts/apps in the local/downstream clusters, performing day2 ops of downstream clusters, creating or deleting clusters or nodes within clusters, rotating Rancher certificates in the local cluster), it is recommended to use the Backup Restore Operator to prevent data loss.
```

moio:
Thanks @jakefhyde - I reworded your piece for consistency with jargon on the k3s and RKE2 manuals and other parts of the document, then copy edited again with some Gemini help for clarity and conciseness as I am not a native English speaker.

I hope this helps

moio:
Terminology comes from official docs:

https://docs.k3s.io/datastore/backup-restore
https://docs.rke2.io/datastore/backup_restore

I did not want to introduce user confusion (although I understand the internal names are clearer for us who work with them every day...)

mallardduck:
Hey @moio / @jakefhyde - I just realized that there are settings in BRO that we use for in-place restores that might make some of this info less accurate. My good pal @inichols is looking into this for us and will follow up for us here once he has confirmation.

inichols:
> @moio This is what I have come up with, feel free to tweak (also @mallardduck feel free to critique as well).
> 
> > In the event of a failure in the local cluster, it may be necessary to restore Rancher to a previous version. There exist two currently supported methods for restoring Rancher: ETCD Restores, and restores performed by Rancher's Backup Restore Operator. ETCD Restores are suitable if no user changes have been made since the snapshot was taken which need to be persisted. For example, if a user takes a snapshot and then makes a change to their local cluster which causes an outage, an ETCD Restore would be the expected restoration mechanism. However, if a user takes a snapshot of the local cluster and then upgrades the kubernetesVersion of the downstream cluster, an ETCD restore would reset the kubernetesVersion and the nodes in this cluster would be running a version of kubernetes that is higher than their spec. Rancher will atempt to reconcile the version, which could have disastrous results. Likewise, if a user were to take a snapshot of the local cluster, rotate all of the nodes in a Rancher provisioned downstream cluster, and then restore from the snapshot, the information on how these nodes were provisioned would be lost, these nodes would be orphaned and in some cases the cluster could become inoperable. As a general rule, when restoring Rancher if any meaningful changes have been made to the local cluster (installing charts/apps in the local/downstream clusters, performing day2 ops of downstream clusters, creating or deleting clusters or nodes within clusters, rotating Rancher certificates in the local cluster), it is recommended to use the Backup Restore Operator to prevent data loss.

Hey, want to provide some information on BRO. It acts similar to an ETCD Restore for the downstream clusters state. If you have a cluster on 1.31.7 and a BRO backup from 1.30.11, a restore will still cause a difference between the running cluster and the Rancher Manager state. This would cause the downstream cluster to go back to 1.30.11. BRO does not take into account for downstream clusters when restoring in place as Prune(the default for in place migrations) removes everything that is not present in the current backup that matches the resourceset that was used for the backup. 

I think this would be a really cool feature though and really set BRO apart from ETCD backups/restores if we could make it downstream cluster aware. However, I think this would need to be a choice during the restore process. Sometimes you would want to completely set a cluster back to how it was during a certain state, but other times it might be okay to keep certain clusters around that were not in the backup.

docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/back-up-restore-usage-guide.md
```diff
@@ -6,6 +6,22 @@ title: Backup Restore Usage Guide
...
+
+You can restore Rancher using two supported methods: cluster Datastore Restores (available for [k3s](https://docs.k3s.io/datastore/backup-restore) and [RKE2](https://docs.rke2.io/datastore/backup_restore) distributions) and Rancher's Backup, as described on this page.
+
+Datastore Restores are suitable when no changes have been made to downstream (managed) clusters since the snapshot. For example, if a local cluster change causes an outage after a snapshot, use Datastore Restore.
+
+However, if you take a Datastore snapshot and then upgrade a downstream cluster's Kubernetes version, a Datastore Restore will revert the downstream cluster's Kubernetes version in the datastore, creating a version mismatch between data in the local cluster and the actual downstream cluster nodes.
+Rancher will then attempt to reconcile the versions, potentially causing unexpected behavior and data loss.
+Similarly, if you take a Datastore snapshot, rotate nodes in a Rancher-provisioned downstream cluster, and then restore from the snapshot, the rotated nodes may become orphaned, possibly making the cluster inoperable. *Recovery from these scenarios is not supported*.
```
moio, line 19: @mallardduck @inichols 

How about the following?

```suggestion
Both methods are suitable when no changes have been made to downstream (managed) clusters since the last snapshot or backup. For example, if a local cluster change such as a version upgrade causes an outage, either a Datastore Restore or a Rancher's backup restore can be used.

However, if you take a snapshot or backup and then upgrade a downstream cluster's Kubernetes version, a subsequent restore will revert the downstream cluster's Kubernetes version in the datastore, creating a version mismatch between data in the local cluster and the actual downstream cluster nodes.
Rancher will then attempt to reconcile the versions, potentially causing unexpected behavior and data loss. Similar considerations apply to other changes to downstream clusters. *Recovery from these scenarios is not supported*.

Rancher's backups are more flexible in those cases, as they allow to selectively restore information from a backup, for example excluding clusters that have been changed since the last backup for manual later recovery.
```

moio:
> Hey, want to provide some information on BRO. It acts similar to an ETCD Restore for the downstream clusters state. If you have a cluster on 1.31.7 and a BRO backup from 1.30.11, a restore will still cause a difference between the running cluster and the Rancher Manager state. This would cause the downstream cluster to go back to 1.30.11. BRO does not take into account for downstream clusters when restoring in place as Prune(the default for in place migrations) removes everything that is not present in the current backup that matches the resourceset that was used for the backup.

I hope I captured that correctly https://github.com/rancher/rancher-docs/pull/1705/files#r2073153167, please let me know

> I think this would be a really cool feature though and really set BRO apart from ETCD backups/restores if we could make it downstream cluster aware. However, I think this would need to be a choice during the restore process. Sometimes you would want to completely set a cluster back to how it was during a certain state, but other times it might be okay to keep certain clusters around that were not in the backup.

I agree, please bring this up to management/PM for further prioritization.

PR #1778, opened 2025-04-30
Fixing a wrong Note header in Vagrant

kakabisht:
## Description

The [note heading in Vagrant Quick Start](https://ranchermanager.docs.rancher.com/getting-started/quick-start-guides/deploy-rancher-manager/vagrant#note) should be an infobox.

I have made the changes in all the versions of the docs.


btat (changes requested):
The original PR (#1776) updated 9 files, but this PR only updates 1. Can you update the other instances too.

btat (approved):

Issue #1797, opened 2025-05-19
Update Logging docs with workaround to change the docker image used when creating a HostTailer resource


diogoasouza:
Due to an [upstream bug](https://github.com/kube-logging/logging-operator/issues/2016), a workaround is needed if a user wants to change the Docker image used by a `HostTailer` resource. 
The goal of this issue is to document the workaround in the [logging docs](https://ranchermanager.docs.rancher.com/integrations-in-rancher/logging).

PR #1798, opened 2025-05-19
add doc on how to use custom image with HostTailer

diogoasouza:
Fixes https://github.com/rancher/rancher-docs/issues/1797

## Description

This PR adds a section in the docs on how to use a custom image when creating a `HostTailer` resource.


diogoasouza:
@pmkovar I think it would be nice to have it in the docs regardless. Usually, people expect it to be `spec.image` but in this case it isn't, so it might save time for some customers

pmkovar:
Thanks @diogoasouza , I think we should update the versioned files as well then?

- https://github.com/rancher/rancher-docs/blob/main/versioned_docs/version-2.11/integrations-in-rancher/logging/logging.md
- https://github.com/rancher/rancher-docs/blob/main/versioned_docs/version-2.10/integrations-in-rancher/logging/logging.md

diogoasouza:
@pmkovar Yes we should. I added a commit with those changes

pmkovar (approved):

PR #1799, opened 2025-05-20
[v2.11.2] Maintenance tasks

pmkovar:
[v2.11.2] Maintenance tasks

Helps #1783.


pmkovar:
Thanks @sunilarjun, good catch. Updated per https://github.com/rancher/rancher/blob/v2.11.2-alpha4/build.yaml to `0.7.2`.

sunilarjun (approved):
Thanks for the quick updates!

PR #1816, opened 2025-06-03
Updating external TLS termination required settings

johnwc:
## Description

Updating external TLS termination required settings. Updating the documentation to be clear on RKE configuration vs RKE2 configuration.

Based on closed issue: https://github.com/rancher/rancher/issues/35088

Fixes [#35088](https://github.com/rancher/rancher/issues/35088)

### DIFF
--- a/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
+++ b/docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
@@ -214,13 +214,30 @@ Your load balancer must support long lived websocket connections and will need t
 
 ### Configuring Ingress for External TLS when Using NGINX v0.25
 
-In NGINX v0.25, the behavior of NGINX has [changed](https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md#0220) regarding forwarding headers and external TLS termination. Therefore, in the scenario that you are using external TLS termination configuration with NGINX v0.25, you must edit the `cluster.yml` to enable the `use-forwarded-headers` option for ingress:
+In NGINX v0.25, the behavior of NGINX has [changed](https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md#0220) regarding forwarding headers and external TLS termination. Therefore, in the scenario that you are using external TLS termination configuration with NGINX v0.25, you must enable the `use-forwarded-headers` option for ingress:
 
+For RKE installations, edit the `cluster.yml` to add the following settings.
 ```yaml
 ingress:
   provider: nginx
   options:
     use-forwarded-headers: 'true'
+```
+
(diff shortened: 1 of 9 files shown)


docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
```diff
@@ -214,15 +214,32 @@ Your load balancer must support long lived websocket connections and will need t
 
 ### Configuring Ingress for External TLS when Using NGINX v0.25
 
-In NGINX v0.25, the behavior of NGINX has [changed](https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md#0220) regarding forwarding headers and external TLS termination. Therefore, in the scenario that you are using external TLS termination configuration with NGINX v0.25, you must edit the `cluster.yml` to enable the `use-forwarded-headers` option for ingress:
+In NGINX v0.25, the behavior of NGINX has [changed](https://github.com/kubernetes/ingress-nginx/blob/master/Changelog.md#0220) regarding forwarding headers and external TLS termination. Therefore, in the scenario that you are using external TLS termination configuration with NGINX v0.25, you must enable the `use-forwarded-headers` option for ingress:
```
pmkovar, line 217:
```suggestion
In NGINX v0.25, the behavior of NGINX has [changed](https://github.com/kubernetes/ingress-nginx/blob/main/Changelog.md?plain=1#L2702) regarding forwarding headers and external TLS termination. Therefore, in the scenario that you are using external TLS termination configuration with NGINX v0.25, you must enable the `use-forwarded-headers` option for ingress:
```

Looks like the changelog is too big to load but we want to refer to v0.22.0, I assume?

pmkovar (changes requested):
Thanks for the PR.

docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
btat: I think there's a good change we can assume v0.25 was a typo. The original link (`Changelog.md#0220`) is formatted for GitHub's preview mode so it would jump to the section header with ID 0220, which ties into Petr's reference to v0.22.0 above. 

In the `Breaking changes` section under v0.22.0 section, there's this [line](https://github.com/kubernetes/ingress-nginx/blob/e4b964bc970d5d16b7a1bce897a98c623fe4b3cd/Changelog.md?plain=1#L2727C2-L2727C153):

>  By default do not trust any client to extract true client IP address from X-Forwarded-For header using realip module (`use-forwarded-headers: "false"`

And this [corresponding PR](https://github.com/kubernetes/ingress-nginx/pull/3333), where the default is change from `true` to `false`.

@pmkovar could you please update your suggestion because of what @johnwc mentioned above about line references not being stable when linking to branches. The example I linked to could be a suitable option.

docs/getting-started/installation-and-upgrade/installation-references/helm-chart-options.md
```diff
@@ -214,15 +214,32 @@ Your load balancer must support long lived websocket connections and will need t
 
 ### Configuring Ingress for External TLS when Using NGINX v0.25
```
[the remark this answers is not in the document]
johnwc: I made all the updates, except for the docs containing Chinese. You'll have to get someone to translate those.

btat (approved):

PR #1841, opened 2025-06-25
Clarify the expected format of `NO_PROXY`

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/50041

## Description

The current documentation doesn't state the expected format of the `NO_PROXY` environment variable. While there are several examples which follow the same format, `NO_PROXY` is not a standardized environment variable and users may be under the impression that various value formats will work, which is not the case. 

For instance, other [SUSE documentation](https://www.suse.com/support/kb/doc/?id=000019675) states that white space can be used as a delimiter. However doing this in Rancher will result in subtle issues that are difficult to debug.

This PR adds a clear explanation of the format of `NO_PROXY` which is accepted by Rancher (and actively tested) to several sections of the documentation that already mention `NO_PROXY`. The explanation includes a link to the official Golang docs which has a full description of the accepted format.

## Comments

Let me know if I should remove the link to the upstream Go documentation in favor of an additional page in the docs repo or some other approach. I'm not 100% clear on this repo's policy on linking to external resources / documentation.

### DIFF
--- a/docs/getting-started/installation-and-upgrade/other-installation-methods/rancher-behind-an-http-proxy/install-kubernetes.md
+++ b/docs/getting-started/installation-and-upgrade/other-installation-methods/rancher-behind-an-http-proxy/install-kubernetes.md
@@ -10,7 +10,15 @@ Once the infrastructure is ready, you can continue with setting up a Kubernetes 
 
 The steps to set up RKE, RKE2, or K3s are shown below.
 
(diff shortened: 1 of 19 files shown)


pmkovar (dismissed):
Thanks for the PR.

I think the upstream link should be totally fine in this repo. Not sure about product docs, @btat do we have a more strict policy there?

PR #1847, opened 2025-06-30
docs: add new audit log docs

alexandreLamarre:
Related to  https://github.com/rancher/rancher/issues/48941

## Reminders

- See the [README](../README.md) for more details on how to work with the Rancher docs.

- Verify if changes pertain to other versions of Rancher. If they do, finalize the edits on one version of the page, then apply the edits to the other versions.

- If the pull request is dependent on an upcoming release, remember to add a "MERGE ON RELEASE" label and set the proper milestone.

## Description

Updates references on:
- How to enable rancher API audit logging
- Documents changes to audit log levels in rancher 2.12+

PR #1948, opened 2025-08-07
[2.12] Add missing flag to enable auditlog

btat:
Related to https://github.com/rancher/rancher-docs/pull/1847.


sunilarjun (approved):
Thanks for catching Billy!

PR #1962, opened 2025-08-20
Updating 2.12 swagger file for 2.12.1 release

sunilarjun:
Related to https://github.com/rancher/rancher-docs/issues/1959.

Generated with [crd-swagger](https://github.com/rancher/crd-swagger) using the command:
`crd-swagger --pretty-print -i docker.io/rancher/rancher:v2.12.1-alpha1 -p 9998 -t 9999 -o openapi.json -f ../rancher/public-api.txt`
 and resource file from https://github.com/rancher/rancher/pull/51048.


btat (changes requested):
`head` shouldn't be used because it represents different images at different times. We should use a static tag instead.

In this case head is for 2.13, which is not what we want.

- https://hub.docker.com/layers/rancher/rancher/head/images/sha256-31bed831bc55da4a50ed69410463aeef0aced58cf06518b044b951a374a675ef
- https://hub.docker.com/layers/rancher/rancher/v2.13-dfd2c793f3addabadcc861f8b89917a2ca15c61a-head/images/sha256-31bed831bc55da4a50ed69410463aeef0aced58cf06518b044b951a374a675ef

sunilarjun:
Closing out as currently no API changes scheduled for 2.12.1.

PR #2150, opened 2026-01-16
Revert "Add FOSSA scanning workflow"

btat:
Reverts rancher/rancher-docs#2146

Not applicable per discussion with the Security team.


sunilarjun (approved):
Thanks @btat!

PR #2180, opened 2026-02-05
Improve air-gapped migration documentation

axeal:
## Description

The purpose of this PR is to improve the Rancher migration documentation (using the rancher-backup operator), to cover use-cases where the environment is fully air-gapped, and the user runs helm commands from a host without internet access. In this instance the user needs to first fetch the charts from a non-air-gapped host, and then copy them to the host with access to the cluster, from which they run the helm commands.

## Comments

I have copied my proposed changes into all versioned documentation from v2.9 through to v2.13. Once the format of the proposed changes here is finalised and merged, I can go ahead and open a similar PR for the rancher-product-docs also.

### DIFF
--- a/docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
+++ b/docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
@@ -48,35 +48,35 @@ Install the [`rancher-backup chart`](https://github.com/rancher/backup-restore-o
      helm install rancher-backup rancher-charts/rancher-backup -n cattle-resources-system --version $CHART_VERSION
      ```
 
-   :::note
+     :::note
 
-   The above assumes an environment with outbound connectivity to Docker Hub.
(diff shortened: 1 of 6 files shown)


docs/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster.md
```diff
@@ -42,23 +42,41 @@ Install the [`rancher-backup chart`](https://github.com/rancher/backup-restore-o
...
   1. Install the charts:
-   
+
      ```bash
      helm install rancher-backup-crd rancher-charts/rancher-backup-crd -n cattle-resources-system --create-namespace --version $CHART_VERSION
      helm install rancher-backup rancher-charts/rancher-backup -n cattle-resources-system --version $CHART_VERSION
      ```
 
-     :::note
+   :::note
```
pmkovar:
```suggestion
     :::note
```

The indentation should be aligned with the previous steps in the procedure, seems broken now.

Same for the following lines, and versioned files.

PR #2223, opened 2026-03-06
Increase max-old-space-size in test deployment

LucasSaintarbor:
Increase max-old-space-size to 8192 bytes in test deployment


btat (changes requested):
The size needs to be updated in package.json as well so that it's taken into account outside of GitHub Actions (e.g. local builds using yarn start and yarn build)

PR #2224, opened 2026-03-06
Updating ns exemptions

sunilarjun:
[SURE-11390] Updating ns exemption list in sample configurations.

Primary source of truth:

https://github.com/rancher/rancher/blob/main/pkg/data/management/podadmissionconfigurationtemplate_data.go


sunilarjun:
Hello @jiaqiluo, when you have a moment do you mind verifying the namespace exemption list to ensure all exemptions necessary are present? This is probably the [best file to review the list](https://github.com/rancher/rancher-docs/pull/2224/files#diff-a523cdb0cb7d8a9f51dd20957ce14b6dec781d25094f294896f8df05ab23eea6) and it is the same update across files just to note. Also please let me know if there is another individual I should reach out to for verification as I went by the `psa-restricted-exemptions.md` file history. Thanks in advance!

PR #2313, opened 2026-05-13
Clarify instructions for restarting cattle-cluster-agent pods after R…

axeal:
…ancher migration

## Reminders

- If the pull request is dependent on an upcoming release, remember to add a "MERGE ON RELEASE" label and set the proper milestone.

## Description

I have updated the language in the Note at the end of the Rancher migration documentation. The primary aim is to remove the phrase `If you wish to keep the original Rancher environment running`, to remove the suggestion that scaling down the original Rancher environment is optional. I still wanted to leave the command documented, in case it is required as a workaround to force reconnection of the cattle-cluster-agents.


pmkovar (approved):
Thanks.

PR #2314, opened 2026-05-13
Add upgrade path requirements for Rancher minor version upgrades

axeal:
## Description

The purpose of this change is to add a clear statement on the required upgrade path between minor versions to the Rancher upgrade documentation. Upgrades should be performed from the latest available patch release of the current minor version to the latest available patch release of the next minor version.


pmkovar (approved):
Thanks.

PR #2315, opened 2026-05-13
Fix helm search command and explicitly call out minor version

axeal:
## Description

Fixes a couple of issues in https://github.com/rancher/rancher-docs/pull/2314 pointed out by a colleague:
- Incorrect `helm search` syntax
- Clarify requirement for minor version upgrades only


pmkovar (approved):
Thanks for the fix.

PR #2343, opened 2026-06-15
Update aws-marketplace to point to SRFA

pmkovar:
Update aws-marketplace to point to SRFA.

cc @jlsarry


LucasSaintarbor (changes requested):
@pmkovar text LGTM but the v2.14 page needs to be updated.

LucasSaintarbor (approved):
@pmkovar Thanks. LGTM.

PR #2360, opened 2026-06-25
Add air-gapped registry note for NeuVector CVE Updater CronJob

allexistence:
Added note about Helm chart registry configuration for air-gapped environments.

## Description

Added a note under **Cluster Support in an Air-Gapped Environment** clarifying that the Helm chart's registry value should point to an internal registry (e.g. Harbor) in air-gapped installs, and that the CVE database Updater CronJob will fail silently if it cannot reach the public NeuVector update endpoint.

### DIFF
--- a/docs/integrations-in-rancher/neuvector/overview.md
+++ b/docs/integrations-in-rancher/neuvector/overview.md
@@ -155,7 +155,11 @@ kubectl patch cronjob neuvector-updater-pod -n cattle-neuvector-system --patch '
 
 :::note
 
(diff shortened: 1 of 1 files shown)


sunilarjun (commented):
Hello @allexistence thank you for the update. Do these changes apply to all versions of the documentation or just latest (v2.14)? If they do need to be applied to other versions you can use the backport script to apply the changes across versions, an example command being `./backport_docs.sh version-2.10 version-2.11 version-2.12 version-2.13 version-2.14` if all supported versions should be updated. Please let me know if there are any questions

